PR #33, opened 2024-03-08
:sparkles: Show blob divert event

foysalit:
This is dependent on https://github.com/bluesky-social/atproto/pull/2238

This PR also adjust label SDK usage within ozone-ui to align with latest `main` on atproto repo.

### DIFF
--- a/components/mod-event/SelectorButton.tsx
+++ b/components/mod-event/SelectorButton.tsx
@@ -101,6 +101,7 @@ export const ModEventSelectorButton = ({
     subjectStatus?.muteUntil,
     subjectStatus?.reviewState,
     subjectStatus?.appealed,
+    hasBlobs,
   ])
   return (
     <Dropdown


components/common/labels/Grid.tsx
```diff
@@ -28,19 +22,10 @@ export const LabelSelector = (props: LabelsProps) => {
...
-        const labelText = typeof label === 'string' ? label : label.id
-        return {
-          label: labelText,
-          value: labelText,
-        }
-      }),
-    }),
-  )
+  const selectorOptions = Object.values(ALL_LABELS).map((labelOption) => ({
+    label: labelOption.identifier,
```
devinivy, line 26: Any idea what this new type error is?

devinivy (approved):
Wasn't positive about the new handling of labels, but all the divert stuff looks good 👍

components/common/labels/Grid.tsx
devinivy: (I assume this is related to @atproto/api updates.)

foysalit: yep, cleaning it up now. I was referencing a local version of the api package so had to adjust things to get the app to run.

bnewbold:
I expected a "Divert" action as an option in the quick-action menu for posts with an image attached, but don't see that in the current render preview?

foysalit:
> I expected a "Divert" action as an option in the quick-action menu for posts with an image attached, but don't see that in the current render preview?

ahh classic useEffect bug. pushed out the fix. mind giving it another go @bnewbold ?

PR #34, opened 2024-03-09
:sparkles: Use labels from 3p labeler

foysalit:


components/common/labels/Grid.tsx
```diff
@@ -21,10 +22,12 @@ export const LabelSelector = (props: LabelsProps) => {
       value: label,
     })),
   )
-  const allOptions = buildAllLabelOptions(defaultLabels, options)
-  const selectorOptions = Object.values(ALL_LABELS).map((labelOption) => ({
-    label: labelOption.identifier,
-    value: labelOption.identifier,
+  const selectorOptions = [
+    ...(client.session?.config.labeler?.policies?.['labelValues'] || []),
+    ...Object.values(ALL_LABELS).map(({ identifier }) => identifier),
```
devinivy, line 27: I wonder if we may find that we'll want to dedupe these.  E.g. in the case that `labelValues` explicitly lists one of the labels in `ALL_LABELS`.

devinivy: And maybe we should place them in a section at top of the combobox?

devinivy:
Replaced by #51

PR #45, opened 2024-03-12
docs: add API overview

bnewbold:
@foysalit did diagrams and most of write-up, I did some copy-editing.


bnewbold:
I'm going to self-merge for now so we have some basic docs.

PR #51, opened 2024-03-13
Show custom labels from labeler service in ozone

foysalit:
https://github.com/bluesky-social/ozone/assets/1919066/633021ba-d46c-44fa-a98e-576defdc3164


foysalit:
commit history is a mess prolly cause I started the branch from divy/distro

PR #52, opened 2024-03-14
Service record configuration

devinivy:
This adds a basic, low-level UI for modifying your labeler's service record using a structured JSON editor.


components/common/Card.tsx
```diff
@@ -1,19 +1,20 @@
...
+  className = '',
   variation = 'default',
-}: {
-  children: React.ReactNode
+  ...props
+}: ComponentProps<'div'> & {
   variation?: Variation
 }) => {
-  let className = 'shadow rounded-sm p-2'
+  className += ' shadow rounded-sm p-2'
```
foysalit, line 12: don't think this needs to be changed but FYI we also have a handy `classNames` util in `@/lib/utils` if you wanna avoid string concatenation like that.

foysalit (approved):
Looks good. wasn't able to run the complete flow against dev-env but I mocked a few responses to simulate some of the states to check and they all worked!

PR #53, opened 2024-03-14
:lipstick: Handle reviewNone, labeler account display and sticky note more prominent

foysalit:
https://github.com/bluesky-social/ozone/assets/1919066/a04134e8-562f-4a4e-a2c3-86d29bf2b595


devinivy (approved):
Great!

PR #61, opened 2024-03-18
Show labeler info in labels and adjust color coding

foysalit:
This PR fetches labeler service details and shows info about the service and the label along with any label on contents to make it easier for moderators to skim through labels.

**NOTE:** As a side effect of behavior based color coding, this PR removes hard coded color coding based on label categories. This is based on the fact that 3rd parties running the service won't benefit from any of the color coding and our own moderators don't seem to rely on the color coding much.

self_label

> Self Label

labeler_service_details

> Labeler service details


app/actions/ModActionPanel/QuickAction.tsx
```diff
@@ -499,7 +501,12 @@ function Form(
...
                     <span className="pl-1">
-                      Last reviewed at:{' '}
+                      Last{' '}
+                      {subjectStatus.reviewState ===
+                      ToolsOzoneModerationDefs.REVIEWNONE
+                        ? 'event'
+                        : 'reviewed'}{' '}
+                      at:{' '}
                       {dateFormatter.format(
                         new Date(subjectStatus.lastReviewedAt),
```
devinivy, line 511: Is `lastReviewedAt` set when the review state is `REVIEWNONE`?

components/common/labels/List.tsx
```diff
@@ -28,3 +43,253 @@ export function LabelChip(props: ComponentProps<'span'>) {
...
+                  </li>
+                ) : null}
+              </ul>
+            )}
+          </>
+        ) : (
+          <h4 className="leading-4">
+            <b>{label.val}</b> label does not have a custom definition. Users
+            will be able to configure the behavior of the label in app.
+          </h4>
```
devinivy, line 290: I'm not 100% sure whether this is accurate—we might want to check in with Paul/Bryan about the right way of messaging around this case.

components/common/labels/useLabelerDefinition.ts
```diff
@@ -0,0 +1,39 @@
...
+      if (labelerDef?.policies.labelValueDefinitions) {
+        const definitionsById: Record<
+          string,
+          ComAtprotoLabelDefs.LabelValueDefinition
+        > = {}
+        labelerDef.policies.labelValueDefinitions.forEach((def) => {
+          definitionsById[def.identifier] = def
+        })
+        labelerDef.policies.definitionById = definitionsById
+      }
```
devinivy, line 28: Is there something in particular that we find in here that we aren't able to get from `session.config.labeler`?

components/common/posts/PostsFeed.tsx
```diff
@@ -418,21 +420,14 @@ function PostLabels({
...
-          <LabelChip
-            className={`${i === 0 ? 'ml-0' : ''} text-[${
-              labelGroup.color
-            }]`}
-            // TODO: Ideally, we should just use inline class name but it only works when the class names are static
-            // so trying to work around that with style prop for now
-            style={{ color: labelGroup.color }}
+          <ModerationLabel
+            recordAuthorDid={item.post.author.did}
+            label={label}
```
devinivy, line 426: I wonder if this needs a `key`.

lib/constants.ts
```diff
@@ -0,0 +1,11 @@
+export const OZONE_SERVICE_DID =
+  process.env.NEXT_PUBLIC_OZONE_SERVICE_DID ||
+  'did:plc:ar7c4by46qjdydhdevvrndac'
```
devinivy, line 3: I am not sure that we need to hardcode our labeler in Ozone.  The service DID may also be set at runtime by fetching `ozone-metadata.json`, as is the case when running the dockerized Ozone distribution.

lib/client.ts
```diff
@@ -132,8 +133,7 @@ class ClientManager extends EventTarget {
   }
 
   private async _getConfig(ozoneDid?: string) {
-    const builtIn =
-      ozoneDid || process.env.NEXT_PUBLIC_OZONE_SERVICE_DID || undefined
+    const builtIn = ozoneDid || OZONE_SERVICE_DID
     return await getConfig(builtIn)
```
devinivy, line 137: For the Ozone distribution we need to leave the possibility open that `builtIn` is not set, and may be determined at runtime within `getConfig`.

devinivy (commented):
The functionality here is awesome!  Just a couple questions for us to answer around runtime configuration and a bit of copy in the app.

lib/constants.ts
foysalit: ooops... accidental commit, will clean up

components/common/labels/useLabelerDefinition.ts
foysalit: so, the idea of this PR is that eventually, we could plug in other labelers into ozone. Bryan touched on this I think somewhere that other labelers may want to see labels added by bluesky. 

there's a bit more work left to get this all to work and get the backend to send labels from other labelers but once that's done, this will make more sense I think.

bnewbold (approved):
this is really great! nice polish

lib/constants.ts
devinivy: @foysalit I think I saw this get cleaned-up, but it still is kicking around here.

components/mod-event/EventItem.tsx
```diff
@@ -152,20 +154,27 @@ const TakedownOrMute = ({
...
+        // Moderation events being displayed means that these events were added by the current service
+        // so we can assume that the src is the same as the configured ozone service DID
         return (
-          <LabelChip key={label} style={{ color: labelGroup.color }}>
-            {displayLabel(label)}
-          </LabelChip>
+          <ModerationLabel
+            key={label}
+            label={{ val: label, src: OZONE_SERVICE_DID, uri: '', cts: '' }}
+          />
```
devinivy, line 177: I don't think we'll be able to capture the "current service" using a constant, since it's determined at runtime.  I think there's a `useSession()` hook kicking around that we could use to grab it, though.

app/actions/ModActionPanel/QuickAction.tsx
```diff
@@ -532,15 +539,14 @@ function Form(
               <FormLabel label="Labels">
                 <LabelList className="-ml-1">
                   {!currentLabels.length && <LabelListEmpty className="ml-1" />}
-                  {currentLabels.map((label) => {
-                    const labelGroup = getLabelGroupInfo(unFlagSelfLabel(label))
+                  {allLabels.map((label) => {
+                    // const labelGroup = getLabelGroupInfo(unFlagSelfLabel(label))
```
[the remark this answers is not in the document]
devinivy: Thought I saw this get cleaned-up too, but it's still here.  Maybe some changes ended-up getting clobbered?

components/mod-event/EventItem.tsx
foysalit: we can also just get it from `client` directly, right?

lib/constants.ts
foysalit: yh idk what's up but I think I lost some changes when merging main into this.

components/common/posts/PostsFeed.tsx
foysalit: there is a key, the last prop.

lib/client.ts
foysalit: yep, cleaning up the fallback of service did to our labeler should be enough for that, no?

app/actions/ModActionPanel/QuickAction.tsx
foysalit: yep.

components/common/labels/List.tsx
foysalit: it's not. some labels are configurable in app though so changed the text from `will be" to "may be" which should have us covered.

devinivy (approved):
Looks good!

PR #63, opened 2024-03-19
:lipstick: Sort email templates alphabetically and allow searching when selecting

foysalit:


app/communication-template/page.tsx
```diff
@@ -56,41 +56,45 @@ export default function CommunicationTemplatePage() {
...
-                  Edit
-                </Link>
-                <ActionButton
-                  appearance="outlined"
-                  onClick={() => setDeletingTemplateId(template.id)}
-                >
-                  <TrashIcon className="h-3 w-3" />
-                </ActionButton>
+        {data
+          ?.sort((prev, next) => prev.name.localeCompare(next.name))
```
devinivy, line 60: This will mutate `data` which may not be desirable to others also holding onto it.  Up above we might do something like:
```ts
const templates = data ?
  [...data].sort((prev, next) => prev.name.localeCompare(next.name)) :
  []
```

And then down here map over `templates`.

devinivy (approved):
Sweet 👍

PR #86, opened 2024-04-05
:bug: Fix url to aturi parser

foysalit:
This PR ensures that at-uri is only built with a DID and not a handle from command palette input.

### DIFF
--- a/components/shell/CommandPalette/useAsyncSearch.tsx
+++ b/components/shell/CommandPalette/useAsyncSearch.tsx
@@ -1,4 +1,3 @@
-import clientManager from '@/lib/client'
 import { getDidFromHandle } from '@/lib/identity'
 import { CollectionId, getCollectionName } from '@/reports/helpers/subject'
 import { AtUri } from '@atproto/api'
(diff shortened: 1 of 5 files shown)


lib/client.ts
```diff
@@ -133,7 +133,10 @@ class ClientManager extends EventTarget {
 
   private async _getConfig(ozoneDid?: string) {
     const builtIn =
-      ozoneDid || process.env.NEXT_PUBLIC_OZONE_SERVICE_DID || undefined
+      ozoneDid ||
+      process.env.NEXT_PUBLIC_OZONE_SERVICE_DID ||
+      'did:plc:kqg2icildynagykzlxehkkaw' ||
```
devinivy, line 138: I think this may be left around from dev.

devinivy (commented):
Looks good!

lib/client.ts
foysalit: yep. cleaned up!

foysalit:
> Looks good! (Any idea what the failing E2E tests are about?)

yep, added some new tests around the changes and the new one was failing. fixed those up just now.

PR #93, opened 2024-04-12
view labels from other labelers in Ozone

foysalit:
Please don't merge/review until the [backend PR](https://github.com/bluesky-social/atproto/pull/2442) has been merged 

https://github.com/bluesky-social/ozone/assets/1919066/fa7a1f2b-474a-40dd-a5d0-9d1fb5390d4c


components/shell/MobileMenu.tsx
```diff
@@ -119,9 +116,6 @@ export function MobileMenu({ toggleTheme }: { toggleTheme: () => void }) {
                   <nav className="flex h-full flex-col">
                     <div className="space-y-1">
                       {NAV_ITEMS.map((item) => {
-                        if (item.serviceAccountOnly && !isServiceAccount) {
```
foysalit, line 122: This check will just need to move to the inside of the page so that the actual config editor is only shown to service accounts.

bnewbold:
I got 5xx errors ("AuthMissing") in Ozone UI when trying to test this in prod.

I think the backend stuff stuff was merged weeks ago so i'd expect to work... maybe an issue with render preview config?

Overall looks good, just wanted to give the actual UX a spin.

components/config/data.ts
```diff
@@ -0,0 +1,27 @@
+import client from '@/lib/client'
+
+const key = 'external_labeler_dids'
```
devinivy, line 3: Distinguishing this as a constant vs. a local variable might be friendly!
```suggestion
const KEY = 'external_labeler_dids'
```

components/config/data.ts
```diff
@@ -0,0 +1,27 @@
...
+  return JSON.parse(labelers)
+}
+
+export const addExternalLabelerDid = (did: string, data: any) => {
+  const labelers = getExternalLabelers()
+  if (labelers[did]) return labelers
+  labelers[did] = data
+  localStorage.setItem(key, JSON.stringify(labelers))
+  return labelers
+}
```
devinivy, line 17: Would be great to type the contents of the local storage items.

components/config/data.ts
foysalit: I often forget my own convention but normally, I try to only uppercase constants if they are being exported but no strong feelings about it.

components/config/data.ts
foysalit: yep yep. I have some helpers set up for these that probably weren't merged when I made this PR. everything should be typed now.

devinivy (approved):
Awesome!

PR #95, opened 2024-04-14
:sparkles: Allow muting reporter

foysalit:
Depends on https://github.com/bluesky-social/atproto/pull/2390

This PR allows moderators to mute reports from a user. It also adds various indicators in the UI around reporter muting such as 
1. A user's review state shows when the user's reports are muted
2. A report event will show a tag if the reporter was muted when it came in

It also adds a way for moderators to view ONLY subjects or reporters that have been muted.


app/actions/ModActionPanel/QuickAction.tsx
```diff
@@ -185,9 +186,15 @@ function Form(
   const isLabelEvent = modEventType === MOD_EVENTS.LABEL
   const isDivertEvent = modEventType === MOD_EVENTS.DIVERT
   const isMuteEvent = modEventType === MOD_EVENTS.MUTE
+  const isUnmuteEvent = modEventType === MOD_EVENTS.UNMUTE
   const isCommentEvent = modEventType === MOD_EVENTS.COMMENT
   const shouldShowDurationInHoursField =
     modEventType === MOD_EVENTS.TAKEDOWN || isMuteEvent
+  const isReportingMutedForSubject = isMuted(subjectStatus, true)
+  const shouldShowMuteReportingOnlyField =
+    isSubjetDid &&
```
devinivy, line 195: Not related to the work in this PR, but I noticed a typo in this variable name: `isSubjetDid` vs. `isSubjectDid`.

app/reports/page-content.tsx
```diff
@@ -126,6 +127,12 @@ const ResolvedFilters = () => {
           onClick: () => updateParams('includeMuted', true),
           isActive: includeMuted === 'true',
         },
+        {
+          id: 'onlyMuted',
+          text: 'Only Muted',
```
devinivy, line 132: What results should you get when "only muted" is on?

components/mod-event/EventItem.tsx
```diff
@@ -82,17 +108,25 @@ const Report = ({
...
+          <LinkToAuthor
+            createdBy={modEvent.createdBy}
+            creatorHandle={modEvent.creatorHandle}
+          />
         </span>
-        {modEvent.event.reportType && !isAppeal && (
-          <ReasonBadge reasonType={modEvent.event.reportType} />
-        )}
-      </p>
+        <div>
```
devinivy, line 119: Not sure what this build issue is 🤔

devinivy (approved):
There may be a few build issues unrelated to @atproto/api typings—aside from that, this is looking good to go! 👍

app/reports/page-content.tsx
foysalit: Only subjects that were muted or reporters that were muted will show up in the queue. This is to provide a quick way to glance through all mutes in case manual reverse/check is necessary.

components/mod-event/EventItem.tsx
foysalit: did you point at the right line? not seeing any issues here? maybe got cleaned up along the way?

PR #110, opened 2024-05-01
✨  Change label selector from searchable dropdown to clickable buttons

foysalit:
- Change the label form to use clickable buttons instead of searchable dropdown. The searchable dropdown package has been removed as dependency. Labels can still be searched and custom labels can be added.
- Add cypress tests around the label add/remove and search behavior.
- Refactor cypress test code to make them more shareable.

Screenshot 2024-05-13 at 21 20 23

> Selectable label buttons

Screenshot 2024-05-13 at 21 20 41

> Search labels and add custom label


components/common/Dropdown.tsx
```diff
@@ -15,6 +15,7 @@ export const Dropdown = ({
   children,
   items,
   containerClassName,
+  ...rest
```
foysalit, line 18: This allows us to pass params like `data-cy` to the wrapper component/

components/mod-event/EventList.tsx
```diff
@@ -375,7 +375,7 @@ const EventFilterPanel = ({
               type="datetime-local"
               id="createdAfter"
               name="createdAfter"
-              className="block w-full"
+              className="block w-full dark:[color-scheme:dark]"
```
foysalit, line 378: The date picker, by default, adds a calendar icon in the input field. This classname ensures that the icon's color changes based on the browser's scheme.

components/mod-event/EventList.tsx
```diff
@@ -478,29 +478,29 @@ const EventFilterPanel = ({
...
-                }),
-              )}
-              onChange={(value) =>
-                changeListFilter({
-                  field: 'reportTypes',
-                  value: Array.isArray(value) ? value.map((v) => v.value) : [],
-                })
-              }
-            />
+            {Object.keys(reasonTypeOptions).map((typeValue) => {
```
foysalit, line 483: Move the report type selector from searchable dropdown to clickable buttons just like labels.

cypress.config.ts
```diff
@@ -1,9 +1,21 @@
-import { defineConfig } from "cypress";
+import { defineConfig } from 'cypress'
 
 export default defineConfig({
   e2e: {
     setupNodeEvents(on, config) {
-      // implement node event listeners here
+      on('before:browser:launch', (browser, launchOptions) => {
```
foysalit, line 6: Make sure the browser window is reasonably wide and tall enough so that tests can run without scrolling too much.

devinivy (approved):
Looks good!

Let's just ensure the lint and updated cypress tests pass here.  Seems like they hung on the first run, so I've restarted them.

PR #111, opened 2024-05-01
:bookmark: Changelog and preparation for new tag

foysalit:
- Update package version for `@atproto/ozone` package to latest
- Update yarn lock file to reflect the new version
- Add a changelog file

### DIFF
--- a/service/package.json
+++ b/service/package.json
@@ -3,7 +3,10 @@
   "description": "Ozone service entrypoint",
   "main": "index.js",
   "dependencies": {
-    "@atproto/ozone": "0.1.1",
+    "@atproto/ozone": "0.1.11",
     "next": "14.0.1"
+  },
+  "volta": {
+    "node": "20.12.2"
   }
 }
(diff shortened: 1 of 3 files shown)


service/package.json
```diff
@@ -3,7 +3,10 @@
   "description": "Ozone service entrypoint",
   "main": "index.js",
   "dependencies": {
-    "@atproto/ozone": "0.1.1",
+    "@atproto/ozone": "0.1.11",
     "next": "14.0.1"
+  },
+  "volta": {
+    "node": "20.12.2"
   }
```
devinivy, line 11: Cool if we nix the volta reference here?  This is primarily for use in the docker build process, hoping to keep it pretty minimal.

foysalit: yh totally, I use it locally and my global node is still at v18 so it was giving me trouble when installing packages using v18.

bnewbold (approved):
changelog lgtm!

PR #113, opened 2024-05-03
:recycle: Refactor the tabs component to be reused

foysalit:
This is a simple refactor of the existing tabs components that was just copy pasted across various pages.

### DIFF
--- a/components/common/Tabs.tsx
+++ b/components/common/Tabs.tsx
@@ -15,15 +15,35 @@ export function Tabs<ViewName>({
   onSetCurrentView: (v: ViewName) => void
   views: TabView<ViewName>[]
 }) {
-  const Tab = ({
-    view,
-    label,
-    sublabel,
-  }: {
-    view: ViewName
-    label: string
-    sublabel?: string
-  }) => (
+  return (
+    <div className="mt-6 sm:mt-2 2xl:mt-5">
+      <div className="border-b border-gray-200">
+        <div className="mx-auto max-w-5xl px-4 sm:px-6 lg:px-8">
+          <nav className="-mb-px flex space-x-8" aria-label="Tabs">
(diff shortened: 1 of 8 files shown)


components/shell/MobileMenu.tsx
```diff
@@ -119,9 +119,6 @@ export function MobileMenu({ toggleTheme }: { toggleTheme: () => void }) {
                   <nav className="flex h-full flex-col">
                     <div className="space-y-1">
                       {NAV_ITEMS.map((item) => {
-                        if (item.serviceAccountOnly && !isServiceAccount) {
-                          return
-                        }
                         const Icon = ICONS[item.icon]
```
devinivy, line 122: Does this end up showing the "Configure" nav item to non-service accounts?  The main issue there is that only the service account has the ability to modify their service record, so it wont have the intended effect if another user does it.

devinivy: (There's also similar logic in SidebarNav.tsx)

components/common/Tabs.tsx
```diff
@@ -0,0 +1,57 @@
...
+}: {
+  currentView: ViewName
+  onSetCurrentView: (v: ViewName) => void
+  views: TabView<ViewName>[]
+}) {
+  const Tab = ({
+    view,
+    label,
+    sublabel,
+  }: {
```
devinivy, line 22: Defining a component anonymously like this has a quirk, which is that it always fully unmounts and remounts on render.  I'd recommend just pulling this out into its own component outside of `Tabs` render function.

components/shell/MobileMenu.tsx
foysalit: oh this wasn't meant to be committed. cleaned up.

devinivy (approved):
Nice!

PR #114, opened 2024-05-06
:sparkles: Show post preview even when moderator is blocked by the author

foysalit:
Since `getPostThread` does not return post record data when there's a block relationship between author and the moderator viewer, this PR builds the post data from `getRecord` so that post preview is displayed properly in the quick action panel.

### DIFF
--- a/components/common/posts/PostsFeed.tsx
+++ b/components/common/posts/PostsFeed.tsx
@@ -222,9 +222,17 @@ function PostEmbeds({ item }: { item: AppBskyFeedDefs.FeedViewPost }) {
     ? item.post.embed.media
     : item.post.embed
 
-  const imageRequiresBlur = doesLabelNeedBlur(
-    item.post.labels?.map(({ val }) => val),
-  )
+  const labelVals: string[] = []
+  if (Array.isArray(item.post.labels)) {
+    item.post.labels.forEach(({ val }) => labelVals.push(val))
+  } else if (Array.isArray(item.post.labels?.['values'])) {
+    // This handles the case where the item may be built from raw record and in that case, labels value is not an array rather a self label object
+    const labels = item.post.labels?.['values'] as unknown as Array<{
+      val: string
+    }>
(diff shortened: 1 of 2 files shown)


components/common/posts/PostsFeed.tsx
```diff
@@ -222,9 +222,17 @@ function PostEmbeds({ item }: { item: AppBskyFeedDefs.FeedViewPost }) {
...
+  const labelVals: string[] = []
+  if (Array.isArray(item.post.labels)) {
+    item.post.labels.forEach(({ val }) => labelVals.push(val))
+  } else if (Array.isArray(item.post.labels?.['values'])) {
+    // This handles the case where the item may be built from raw record and in that case, labels value is not an array rather a self label object
+    const labels = item.post.labels?.['values'] as unknown as Array<{
+      val: string
+    }>
+    labels.forEach(({ val }) => labelVals.push(val))
+  }
```
devinivy, line 234: Rather than passing a raw record here and deviating in this way from the expected, valid `FeedViewPost` type, I wonder if it may be friendlier to instead build the `labels` object on the other side and avoid this special case on the consumer side.

Might look something like:
```ts
  if (data?.thread?.blocked) {
    return (
      <BaseRecordCard
        uri={uri}
        renderRecord={(record) => (
          <PostAsCard
            dense
            controls={false}
            item={{
              post: {
                uri: record.uri,
                cid: record.cid,
                author: record.repo,
                record: record.value,
                labels: isSelfLabels(record.value['labels'])
                  ? record.value['labels'].values.map(({ val }) => ({
                      val,
                      uri: record.uri,
                      src: record.repo.did,
                      cts: new Date(0).toISOString(),
                    }))
                  : [],
                indexedAt: new Date(0).toISOString(),
              },
            }}
          />
        )}
      />
    )
  }
```

foysalit: yeah that looks cleaner, thanks!

devinivy (approved):

PR #115, opened 2024-05-07
❇️ Moderator user manager

foysalit:
Depends on https://github.com/bluesky-social/atproto/pull/2460

This PR allows user management from ozone UI.

https://github.com/bluesky-social/ozone/assets/1919066/6f8ebbd4-80ee-4851-a3b0-c31eb59afa51

> FIrst look - UI might change


components/config/Member.tsx
```diff
@@ -0,0 +1,58 @@
...
+    <div className="pt-4">
+      <div className="flex flex-row justify-between mb-4">
+        <h4 className="font-medium text-gray-700 dark:text-gray-100">
+          Manage Members
+        </h4>
+        {!showMemberCreateForm && !editingMember && (
+          <ActionButton
+            size="sm"
+            appearance="primary"
+            disabled={!!editingMember}
```
devinivy, line 32: Looks like the member is never being edited here, so `disabled` would always be `false`.
```suggestion
        {!showMemberCreateForm && !editingMember && (
          <ActionButton
            size="sm"
            appearance="primary"
```

components/team/MemberEditor.tsx
```diff
@@ -0,0 +1,209 @@
...
+            error: `Could not resolve handle ${didOrHandle}`,
+          })
+          return
+        }
+        did = didFromHandle
+      }
+
+      // Normally we wouldn't use <any> but the result of the request does not change
+      // anything in the UI so we don't need to type it
+      let request: Promise<any>
```
devinivy, line 53: The `unknown` type seems like it may be suitable here, and safer/friendlier from a typing perspective.
```suggestion
      // The result of the request does not change anything in the UI so the type can be unknown
      let request: Promise<unknown>
```

components/team/MemberEditor.tsx
```diff
@@ -0,0 +1,209 @@
...
+            appearance="primary"
+            disabled={submission.isSubmitting}
+          >
+            {!submission.isSubmitting
+              ? !!member
+                ? 'Update Member'
+                : 'Add Member'
+              : !!member
+              ? 'Updating Member...'
+              : 'Adding Member...'}
```
devinivy, line 192: This one is a bit tricky to read through for me!

devinivy (approved):
Ace!

components/team/MemberEditor.tsx
foysalit: yeah went a bit too crazy with the ternaries lol. refactored it into its own helper

components/config/Member.tsx
foysalit: oh right I decided to just hide the button when in edit mode.

PR #120, opened 2024-05-21
:sparkles: Use q instead of term for searcning repos

foysalit:
`term` param has been deprecated in favor of `q` https://github.com/bluesky-social/atproto/pull/2507


devinivy (approved):

bnewbold (approved):
thanks for the quick update!

PR #123, opened 2024-05-22
✨ Config based UI

foysalit:
Use the config published from ozone server [here](https://github.com/bluesky-social/atproto/pull/2494) to only show relevant content based on viewer's role.

Also, display the config in the configuration page.

Screenshot 2024-05-22 at 23 45 13

> Dark Mode

Screenshot 2024-05-22 at 23 45 21

> Light Mode


lib/server-config.ts
```diff
@@ -0,0 +1,39 @@
...
+    pds: config.pds?.url,
+    blobDivert: config.blobDivert?.url,
+    appview: config.appview?.url,
+    chat: config.chat?.url,
+    permissions: {
+      canManageTemplates: isModerator,
+      canTakedown: !!config.pds?.url && isModerator,
+      canLabel: isModerator,
+      canManageChat: !!config.chat?.url && isModerator,
+      canSendEmail: !!config.pds?.url && isModerator,
```
devinivy, line 36: Just double-checking that there aren't any admin-only actions.

devinivy (approved):
Lovely

lib/server-config.ts
foysalit: there are a couple of admin only things like managing team members (which we just merged) and feedgen takedown and I just added those 2 in.

PR #128, opened 2024-05-31
OAuth integration

matthieusieben:
Running locally:
1) Run `pnpm i && pnpm build && pnpm dev` from `atproto`'s [feat-oauth-client](https://github.com/bluesky-social/atproto/tree/feat-oauth-client) branch
2) Copy the "🗼 Ozone service DID" value printed in the outpur
2) From this branch: `NEXT_PUBLIC_PLC_DIRECTORY_URL="http://localhost:2582" NEXT_PUBLIC_SOCIAL_APP_URL="http://localhost:2584" NEXT_PUBLIC_OZONE_SERVICE_DID="<ENTER_COPIED_OZONE_DID_HERE>" y dev`
4) Visit `http://127.0.0.1:3000` (make sure to use `127.0.0.1` and **not** `localhost`). Alternatively, start a reverse proxy using: `ngrok http 3000` and visit the proxy's URL instead


foysalit (commented):
Leaving some comments after reviewing nearly 60% of the changes. Will do another pass once you get the rebased changes in.
Again, apologies for the huge conflict you're handling right now and LMK if I can help!

foysalit (commented):
wow this is a massive piece of work! thank you for powering through the conflict resolution and it looks quite good. left some comments and questions but nothing major popped out. 

would love to get an extra pair of eyes from @devinivy and then do some testing with our mod team before creating a distro with this so let's hold on to merging until we get a distro out this week. will come back and approve once that's done.

app/actions/ModActionPanel/QuickAction.tsx
```diff
@@ -159,6 +163,10 @@ function Form(
     replaceFormWithEvents: boolean
   } & Pick<Props, 'setSubject' | 'subject' | 'subjectOptions' | 'onSubmit'>,
 ) {
+  const queryClient = useQueryClient()
+  const labelerAgent = useLabelerAgent()
+  const accountDid = labelerAgent.getDid()
```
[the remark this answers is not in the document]
matthieusieben: Yeah you need to work with the `feat-oauth-client` branch from the `atproto` repo. Requires some tweaks to work locally. I can explain how I do it if you need.

components/shell/ConfigurationContext.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+export function useLabelerAgent() {
+  return useConfigurationContext().labelerAgent
+}
+
+export function useServerConfig() {
+  return useConfigurationContext().serverConfig
+}
+
+export function usePermission(name: PermissionName) {
```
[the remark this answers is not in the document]
matthieusieben:
```ts 
const canFoo = usePermission('foo')
if (canFoo) {
  // OK
}
```

This is short hand for
```ts
const {serverConfig} = useConfigurationContext()
if (serverConfig.permissions["foo"]) {
  // OK
}
```

Both are exactly identical.

components/config/server-config.tsx
```diff
@@ -10,39 +13,26 @@ import {
...
     >
-      {isRefetching ? 'Refetching...' : 'Refetch'}
+      {updateRecord.isLoading ? 'Refetching...' : 'Refetch'}
     </ActionButton>
   )
 }
 
-export const ServerConfig = ({ session }: { session: ClientSession }) => {
-  const config = session.serverConfig
-  if (!config) {
```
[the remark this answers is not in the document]
matthieusieben: `serverConfig` was completely split out from session management indeed.

If you look [here](https://github.com/bluesky-social/ozone/blob/oauth-matthieu/app/layout.tsx#L52-L67) you'll see that the `AuthProvider` wraps the `ConfigurationProvider`, and [here](https://github.com/bluesky-social/ozone/blob/oauth-matthieu/components/shell/ConfigurationContext.tsx#L88-L107) will show that the `serverConfig` will actually be loaded after auth was successful (`labelerAgent` available), and does not need to have been stored in the session at all.

app/layout.tsx
```diff
@@ -45,13 +48,23 @@ export default function RootLayout({
...
-        <AuthProvider>
-          <CommandPaletteRoot>
-            <QueryClientProvider client={queryClient}>
-              <Shell>{children}</Shell>
-            </QueryClientProvider>
-          </CommandPaletteRoot>
-        </AuthProvider>
+
+        <GlobalQueryClientProvider>
+          <AuthProvider
```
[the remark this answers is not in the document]
matthieusieben: That's because the package was not released on NPM, meaning that types do not resolve properly in CI. Types are good locally ;-)

app/repositories/[id]/[...record]/page-content.tsx
```diff
@@ -103,7 +106,7 @@ export default function RecordViewPageContent({
           return undefined
         }
         // TODO: We need pagination here, right? how come getPostThread doesn't need it?
-        const { data: listData } = await client.api.app.bsky.graph.getList({
+        const { data: listData } = await pdsAgent.api.app.bsky.graph.getList({
```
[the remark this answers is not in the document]
Comment (matthieusieben) [reply app/repositories/[id]/[...record]/page-content.tsx:109]: You tell me. I kept the current behavior (that did not include the proxy headers):

https://github.com/bluesky-social/ozone/blob/134524d857a507600c002c6b14a8b51fbe06cc42/app/repositories/%5Bid%5D/%5B...record%5D/page-content.tsx#L106-L108

components/common/Tabs.tsx
```diff
@@ -66,3 +67,34 @@ function Tab<ViewName>({
     </span>
   )
 }
+
+export function TabsPanel<ViewName extends Key>({
```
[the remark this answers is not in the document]
matthieusieben: I added prop to allow manipulating the current view from outside, allowing the parent component to sync the tab with the query params (if it wants to).

I did not "bake" this behavior here though a I believe this component should not concern itself with routing stuff for proper separation of concerns. wdyt?

components/list/Lists.tsx
```diff
@@ -1,30 +1,27 @@
...
+        limit: 25,
+        cursor: pageParam,
+      })
+      return data
+    },
+    getNextPageParam: (lastPage) => lastPage.cursor,
+  })
 
-  if (isInitialLoading) {
+  if (isLoading) {
```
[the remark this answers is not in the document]
matthieusieben: wrong ;-)
Capture d’écran 2024-07-09 à 11 10 05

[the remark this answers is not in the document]
matthieusieben: https://tanstack.com/query/latest/docs/framework/react/reference/useQuery#usequery

components/shell/AuthContext.tsx
```diff
@@ -1,50 +1,153 @@
-import { AuthState } from '@/lib/types'
-import { createContext, useState } from 'react'
+'use client'
 
-// This may seem a bit over-engineered but this is only the groundwork so while the current data model is simplistic
-// and could probably be handled in a simpler setup, the below setup helps us evolve the context to hold much complex
-// model and expose only what may be needed by UI if/when we have complex authorization based on user role etc.
+import { AppBskyActorDefs, BskyAgent } from '@atproto/api'
+import { isLoopbackHost } from '@atproto/oauth-client-browser'
```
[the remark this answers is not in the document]
matthieusieben: Actually not deployed on NPM yet. This only works locally until `feat-oauth-client` branch from `atproto` repo is merged & published.

foysalit (approved):
Great work!

matthieusieben:
Hey @foysalit, thanks for fixing this branch !

Note that a small [refactor](https://github.com/bluesky-social/atproto/pull/2714) due to a design change we made will break this branch once landed. The fix is trivial though.

middleware.ts
```diff
@@ -1,17 +0,0 @@
...
-}
-
-export function middleware(request: NextRequest) {
-  const url = request.nextUrl.clone()
-  if (url.pathname === '/') {
-    url.pathname = '/reports'
-    url.searchParams.set('resolved', 'false')
-    return NextResponse.redirect(url)
-  }
-}
```
devinivy, line 17: Why do we no longer need this middleware?

lib/constants.ts
```diff
@@ -1,13 +1,32 @@
...
+  process.env.NEXT_PUBLIC_SOCIAL_APP_URL ||
+  (process.env.NODE_ENV === 'development'
+    ? 'http://localhost:2584'
+    : 'https://bsky.app')
+
+export const HANDLE_RESOLVER_URL =
+  process.env.NEXT_PUBLIC_HANDLE_RESOLVER_URL ||
+  (process.env.NODE_ENV === 'development'
+    ? 'http://localhost:2584'
+    : 'https://bsky.social')
```
devinivy, line 29: The endpoint on the appview is going to be more reliable for this purpose.  Perhaps we can eventually implement this endpoint on the Ozone backend and it can serve itself.
```suggestion
export const HANDLE_RESOLVER_URL =
  process.env.NEXT_PUBLIC_HANDLE_RESOLVER_URL ||
  (process.env.NODE_ENV === 'development'
    ? 'http://localhost:2584'
    : 'https://api.bsky.app')
```

environment.d.ts
```diff
@@ -6,6 +6,8 @@ declare global {
       NEXT_PUBLIC_PLC_DIRECTORY_URL?: string // e.g. https://plc.directory
       NEXT_PUBLIC_QUEUE_CONFIG?: string
       NEXT_PUBLIC_OZONE_SERVICE_DID?: string // e.g. did:plc:xxx#atproto_labeler
+      NEXT_PUBLIC_OZONE_PUBLIC_URL?: string // e.g. https://ozone.example.com (falls back to window.location.origin)
+      NEXT_PUBLIC_SOCIAL_APP_URL?: string // e.g. https://bsky.app
     }
```
devinivy, line 11: I believe I saw an additional env related to handle resolution.

cypress/support/api.ts
```diff
@@ -1,6 +1,7 @@
-export const API_URL = 'https://bsky.social'
+export const API_URL = 'http://localhost:2583'
 export const SERVER_URL = `${API_URL}/xrpc`
 export const PLC_URL = 'https://plc.directory'
+export const HANDLE_RESOLVER_URL = 'https://bsky.social/xrpc'
```
devinivy, line 4: I recommend using the appview for handle resolution:
```suggestion
export const HANDLE_RESOLVER_URL = 'https://api.bsky.app/xrpc'
```

components/shell/ConfigurationFlow.tsx
```diff
@@ -1,59 +1,112 @@
...
+  }
+
+  if (state === ConfigurationState.Unauthorized) {
     return (
       <>
         <ErrorInfo type="warn" className="mt-2">
-          {`You're`} not logged-in. Please login using your Ozone service
-          account in order to configure Ozone.
+          Your credentials are not authorized on this network. Please login as
+          an authorized user in order to configure Ozone.
```
devinivy, line 56: We specifically want the user to login using their service account, i.e. the account with the same DID as their labeler.  I don't think we need to iterate on the verbiage of the configuration flow as part of the OAuth integration, unless there's some new auth state that needs to be incorporated into the configuration flow.

devinivy:
```suggestion
          {`You're`} not logged-in. Please login using your Ozone service
          account in order to configure Ozone.
```

app/oauth-client.json/route.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  return Response.json(
+    oauthClientMetadataSchema.parse({
+      client_id: clientMetadataUrl.href,
+      client_uri: clientMetadataUrl.href,
+      redirect_uris: [new URL('/', clientMetadataUrl).href],
+      response_types: ['code id_token', 'code'],
+      grant_types: ['authorization_code', 'implicit'],
+      token_endpoint_auth_method: 'none',
+      scope: 'openid profile email',
```
devinivy, line 25:
```suggestion
      response_types: ['code'],
      grant_types: ['authorization_code', 'refresh_token'],
      token_endpoint_auth_method: 'none',
      scope: 'atproto transition:generic',
```

components/shell/CommandPalette/useAsyncSearch.tsx
```diff
@@ -299,7 +299,17 @@ export const useCommandPaletteAsyncSearch = () => {
...
+                return router.push(`?quickOpen=${atUri}`)
+              }
+
+              const { handle } = fragments
+              if (handle) {
+                getDidFromHandle(handle).then((did) => {
+                  if (!did) return
+                  router.push(`?quickOpen=${atUri.replace(handle, did)}`)
+                })
+              }
```
devinivy, line 312: Is this related to the OAuth integration, does it fix a bug, or a misc change/improvement?

components/shell/auth/oauth/useOAuth.ts
```diff
@@ -0,0 +1,288 @@
...
+              } else {
+                client.dispose()
+              }
+            },
+            (err) => {
+              if (!signal.aborted) throw err
+            },
+          )
+        }
+      })
```
devinivy, line 132: This gets a bit complicated to handle the branching paths of loading client metadata vs. baking it in.  I assume we just use one of these paths in ozone— can we remove the other?

lib/client-config.ts
```diff
@@ -1,37 +1,41 @@
...
   }
-  labelerDid ??= meta?.did
-  if (!labelerDid) {
+
+  const did = meta?.did ?? labelerDid
+
+  if (!did) {
     throw new Error('Could not determine an Ozone service DID')
+  } else if (labelerDid && did !== labelerDid) {
+    throw new Error('Mismatch between Ozone service DID and meta info')
```
devinivy, line 29:
```suggestion
    throw new Error('Mismatch between Ozone service DID and Ozone service metadata')
```

middleware.ts
devinivy: Ah, found the client-side redirect 👍

devinivy (approved):

components/shell/auth/oauth/useOAuth.ts
matthieusieben: Actually did that without even seeing your comment here 😅

middleware.ts
matthieusieben: The redirect uri is `/`, which is conflicting with the logic here.

Also, the oauth flow uses the `state` param to "remember" the page that was being visited (in case credentials are no longer valid). This mechanism also conflicts with this middleware.

components/shell/CommandPalette/useAsyncSearch.tsx
matthieusieben: @foysalit , I believe this was you ?

foysalit: yep, this is a minor bugfix that I noticed while testing this branch so just put it in here. 
essentially, at-uris can contain handle instead of DID so this just makes sure that we resolve the handle to DID before doing the reroute since ozone always wants to use DIDs in URL.

lib/constants.ts
```diff
@@ -1,13 +1,30 @@
...
 export const OZONE_SERVICE_DID =
   process.env.NEXT_PUBLIC_OZONE_SERVICE_DID || undefined
 
+export const OZONE_PUBLIC_URL =
+  process.env.NEXT_PUBLIC_OZONE_PUBLIC_URL || undefined
+
 export const PLC_DIRECTORY_URL =
-  process.env.NEXT_PUBLIC_PLC_DIRECTORY_URL || `https://plc.directory`
+  process.env.NEXT_PUBLIC_PLC_DIRECTORY_URL ||
+  (process.env.NODE_ENV === 'development'
```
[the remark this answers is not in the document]
foysalit: to add more context, it makes total sense that when developing locally, we'd want to point to local atproto dev env but often, I point at staging or production to check UI things which is adds extra step when committing changes etc. so, this is more of a devex consideration.

foysalit (approved):

lib/constants.ts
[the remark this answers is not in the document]
matthieusieben: I added this so that it works with `yarn dev` without any additional `.env` file whatsoever. I am happy to remove this if that breaks your flow.

lib/client-config.ts
```diff
@@ -9,10 +9,16 @@ export async function getConfig(): Promise<OzoneConfig> {
   if (labelerDid) {
     doc = await resolveDidDocData(labelerDid)
     const labelerUrl = doc && getServiceUrlFromDoc(doc, 'atproto_labeler')
-    if (labelerUrl) {
-      meta = await getOzoneMeta(labelerUrl)
+    if (process.env.NODE_ENV === 'development' && doc && labelerUrl) {
```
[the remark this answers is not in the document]
matthieusieben: This is solely for the purpose of making this work with `yarn dev`

PR #129, opened 2024-06-03
:sparkles: Show account deactivation flag in action panel and account page

foysalit:
Screenshot 2024-06-03 at 19 35 16

> On post from deactivated account

Screenshot 2024-06-03 at 19 26 55

> On account page

### DIFF
--- a/lib/client.ts
+++ b/lib/client.ts
@@ -84,7 +84,7 @@ class ClientManager extends EventTarget {
       refreshJwt: login.refreshJwt,
       handle: login.handle,
       did: login.did,
-      active: !!login.active,
+      active: login.active !== false,
     }
     this._agent = agent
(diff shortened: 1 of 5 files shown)


lib/client.ts
```diff
@@ -89,6 +84,7 @@ class ClientManager extends EventTarget {
       refreshJwt: login.refreshJwt,
       handle: login.handle,
       did: login.did,
+      active: !!login.active,
```
devinivy, line 87: If `active` is missing then the account most likely is active.  It makes the most sense when you look at it from a compatibility perspective—this is a new field and before the field existed the accounts were assumed active.
```suggestion
      active: login.active !== false,
```

devinivy (approved):

PR #130, opened 2024-06-06
:sparkles: Show embedded record in message context view

foysalit:
Screenshot 2024-06-06 at 16 02 39

### DIFF
--- a/components/common/posts/PostsFeed.tsx
+++ b/components/common/posts/PostsFeed.tsx
@@ -216,8 +216,7 @@ function ImageAltText({ alt }: { alt: string }) {
   )
 }
 
-// @TODO record embeds
-function PostEmbeds({ item }: { item: AppBskyFeedDefs.FeedViewPost }) {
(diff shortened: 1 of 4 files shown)


components/common/posts/PostsFeed.tsx
```diff
@@ -216,8 +216,7 @@ function ImageAltText({ alt }: { alt: string }) {
   )
 }
 
-// @TODO record embeds
-function PostEmbeds({ item }: { item: AppBskyFeedDefs.FeedViewPost }) {
```
foysalit, line 220: This diff looks super complex but the overview is that we just moved out the rendering for embed record into it's own component so that it can be used inside message context viewer.

devinivy (approved):
There are a couple type errors that got flagged from the @atproto/api version upgrade, but otherwise looking good!

foysalit:
> There are a couple type errors that got flagged from the @atproto/api version upgrade, but otherwise looking good!

yep just needed to sync with main

PR #132, opened 2024-06-12
✨ Filter macros

foysalit:
This PR adds a new feature that allows users to save event filter in their browser. Saved filters can be easily applied with one click to easily acquire a smaller subset of events in any place the entire event log is being displayed.

https://github.com/bluesky-social/ozone/assets/1919066/0df2da13-6430-41d9-893b-fc1d9758070b

### DIFF
--- a/app/events/filters/macros/page.tsx
+++ b/app/events/filters/macros/page.tsx
@@ -47,7 +47,7 @@ export default function EventFiltersMacrosListPage() {
           Event Filter Macros
         </h2>
         <LinkButton
-          href="/communication-template/create"
+          href="/events"
           appearance="primary"
           size="sm"
         >
--- a/app/events/filters/macros/page.tsx
+++ b/app/events/filters/macros/page.tsx
@@ -1,0 +1,109 @@
+'use client'
+import { useTitle } from 'react-use'
+import { Loading, LoadingFailed } from '@/common/Loader'
+import {
+  useFilterMacroList,
+  useFilterMacroRemoveMutation,
+} from '@/mod-event/useFilterMacrosList'
+import { Card } from '@/common/Card'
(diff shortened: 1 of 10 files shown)


components/mod-event/useFilterMacrosList.tsx
```diff
@@ -0,0 +1,78 @@
...
+  getList,
+  removeFromList,
+  updateList,
+} from './helpers/macros'
+import { EventListState } from './useModEventList'
+
+const FILTER_MACRO_LIST_QUERY_KEY = 'filter-macro-list'
+
+export const useFilterMacroList = () => {
+  return useQuery({
```
devinivy, line 15: Nice/fun use of react-query in here!

app/events/filters/macros/page.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const [isOpen, setIsOpen] = useState(false)
+  const { data: macroList, isFetching, error } = useFilterMacroList()
+  const {
+    mutateAsync: removeMacro,
+    error: removeMacroError,
+    isLoading: isRemovingMacro,
+  } = useFilterMacroRemoveMutation()
+
+  let pageTitle = `Moderation Filter Macros`
+  useTitle(pageTitle)
```
devinivy, line 30: Any issue with this?
```suggestion
  useTitle('Moderation Filter Macros')
```

devinivy (approved):
Wheew!  This is gonna rule.

app/events/filters/macros/page.tsx
foysalit: no issues, just built up a habbit of doing this so that if and when we do end up making the title dynamic because of whatever reason, it's 1 less line of change lo.

PR #133, opened 2024-06-13
Update HOSTING.md

kristojorg:
This is a tiny update. It was unclear to me what `ozone@example.com` should be replaced with. In my case with the domain `ozone.procycling.social`, replacing it with `ozone@procycling.social` worked, but I'm not sure what it should be replaced with if your domain is a root domain for example.


bnewbold:
We should probably have that line read:

> Be sure to replace `ozone.example.com` with your own domain, and `ozone@example.com` with a technical contact email address (for Let's Encrypt TLS certificates).

kristojorg:
Ah, so it's just a technical contact email? When I tried to run this with the default ozone@example.com, the app wouldn't start (although I don't remember the exact error right now). I will update the line in my PR in any case

PR #134, opened 2024-06-14
:sparkles: Add modal image viewer

foysalit:
This PR introduces a lightbox plugin to open images within a modal instead of opening them in a new tab. This allows moderators to view and navigate images in full size.
The plugin being used is https://yet-another-react-lightbox.com/

While testing, I also found a few UI issues with text color being off in some places so those were fixed too.

Screenshot 2024-06-14 at 21 52 55

> Screenshot from a post's image being opened in lightbox showing thumbnails and alt text as caption

### DIFF
--- a/components/common/posts/ImageList.tsx
+++ b/components/common/posts/ImageList.tsx
@@ -1,0 +1,53 @@
+import { AppBskyEmbedImages } from '@atproto/api'
+import { useState } from 'react'
+import { InformationCircleIcon } from '@heroicons/react/24/outline'
+import Lightbox from 'yet-another-react-lightbox'
+import Captions from 'yet-another-react-lightbox/plugins/captions'
+import Thumbnails from 'yet-another-react-lightbox/plugins/thumbnails'
+import 'yet-another-react-lightbox/styles.css'
+import 'yet-another-react-lightbox/plugins/thumbnails.css'
+import 'yet-another-react-lightbox/plugins/captions.css'
+
+function ImageAltText({ alt }: { alt: string }) {
+  if (!alt) return null
+  return (
+    <p className="leading-2 text-gray-400 text-xs leading-3 mt-1">
+      <InformationCircleIcon className="w-4 h-4 inline mr-1" />
+      {alt}
+    </p>
+  )
+}
+
+export const ImageList = ({
+  images,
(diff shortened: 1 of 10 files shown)


bnewbold (approved):
gave it a try, works as expected!

components/common/posts/ImageList.tsx
```diff
@@ -0,0 +1,53 @@
...
+        plugins={[Thumbnails, Captions]}
+        carousel={{ finite: true }}
+        close={() => setIsImageViewerOpen(false)}
+        slides={images.map((img) => ({
+          src: img.fullsize,
+          description: img.alt,
+        }))}
+      />
+      {images.map((image, i) => (
+        <figure key={`img-${i}`}>
```
devinivy, line 42: This works just as well!
```suggestion
        <figure key={i}>
```

components/common/posts/ImageList.tsx
```diff
@@ -0,0 +1,53 @@
+import { AppBskyEmbedImages } from '@atproto/api'
+import { useState } from 'react'
+import { InformationCircleIcon } from '@heroicons/react/24/outline'
+import Lightbox from 'yet-another-react-lightbox'
+import Captions from 'yet-another-react-lightbox/plugins/captions'
+import Thumbnails from 'yet-another-react-lightbox/plugins/thumbnails'
+import 'yet-another-react-lightbox/styles.css'
+import 'yet-another-react-lightbox/plugins/thumbnails.css'
+import 'yet-another-react-lightbox/plugins/captions.css'
```
devinivy, line 9: I see these CSS imports a couple places.  Can/should we hoist them up near the entrypoint rather than pull them in wherever `yet-another-react-lightbox` is used?

devinivy (approved):
Nice!

components/common/posts/ImageList.tsx
foysalit: oh the other places I included it were leftovers from initial implementation. I was thinking including it in the component would be better so that we are not loading these on initial landing and once you visit a screen where the component is rendered, the css would be cached and future requires would just get cache hit.
however, these are very lightweight, like 3kb each so not too worried about it the load time either way. ended up including it in layout file.

PR #135, opened 2024-06-18
:sparkles: Add Repost, quote, media and no reply filters to author feed

foysalit:
This PR adds new filters for moderators to view specific types of posts from a user, such as excluding replies, showing media only, reposts only, and quotes only posts. Some of the filtering happens client-side and mutually exclusive with the existing keyword search due to differing data sources.

https://github.com/bluesky-social/ozone/assets/1919066/1242e2a3-4951-4c6f-9dd7-b7d83cf3a7be

### DIFF
--- a/components/common/feeds/AuthorFeed.tsx
+++ b/components/common/feeds/AuthorFeed.tsx
@@ -4,6 +4,66 @@ import { Posts } from '../posts/Posts'
 import client from '@/lib/client'
 import { useState } from 'react'
 import { useRepoAndProfile } from '@/repositories/useRepoAndProfile'
+import { AppBskyFeedDefs } from '@atproto/api'
+import { TypeFilterKey, TypeFiltersByKey } from '../posts/constants'
+
+const getAuthorFeed = async ({ id, pageParam, typeFilter, options }) => {
+  const limit = 30
+  let filteredFeed: AppBskyFeedDefs.FeedViewPost[] = []
+  let cursor = pageParam
+  const isPostFilter = [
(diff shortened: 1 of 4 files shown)


bnewbold:
from watching the video: nice! I can imagine in the long-run there being friction or trouble with the way there are separate implementations for keyword search vs post-type-filtering, but I think this is a solid first iteration and will be useful for most cases.

components/common/feeds/AuthorFeed.tsx
```diff
@@ -4,6 +4,66 @@ import { Posts } from '../posts/Posts'
...
+      }
+      return isRepost || isQuotePost
+    })
+
+    filteredFeed = [...filteredFeed, ...newFilteredItems]
+
+    // If no more items are available, break the loop to prevent infinite requests
+    if (!data.cursor) {
+      break
+    }
```
devinivy, line 59: Truly no real issue here, but I often find myself reaching for this pattern when paginating which handles this control flow:
```ts
let cursor: string | undefined
do {
  const page = await getPage(cursor)
  cursor = page.cursor
} while (cursor)
```

devinivy (approved):
Great!

components/common/feeds/AuthorFeed.tsx
foysalit: this is a tiny bit different though, we don't want to iterate through all pages, we only want to iterate until the original page size is reached but before that, if we go through all pages already, we want to break even if the page size is not reached. we could combine the conditions but it might be a bit more confusing that way.

PR #136, opened 2024-06-18
:sparkles: Show search results when searching by email: prefix

foysalit:


app/repositories/page-content.tsx
```diff
@@ -4,22 +4,79 @@ import { useSearchParams } from 'next/navigation'
...
+      // Set placeholder properties that will be later filled in with data from ozone
+      relatedRecords: [],
+      indexedAt: account.indexedAt,
+      moderation: {},
+      labels: [],
+    }
+  })
+
+  await Promise.allSettled(
+    data.accounts.map(async (account) => {
```
devinivy, line 61: Without handling the result of `allSettled()` this will swallow errors, including unexpected ones.  Another option is to use `Promise.all() here` and handle expected error conditions within each item.

devinivy (approved):
Looking good

app/repositories/page-content.tsx
foysalit: This is intentional, I think we have most info we need to render a partial view from the resultSet so if some requests fail, they will just not show the augmented data from getRepo()

devinivy: That totally makes sense and I don't think we'd need to change that— my main question is whether we want to swallow all errors versus ignore some acceptable set of errors and log or throw the rest.  For example, a logic bug could cause a `TypeError` in here and there's no way we'd be able to observe it, could easily go unnoticed.

devinivy: The rule here might be that we're comfy ignoring all `XRPCErrors`, for example.  I really appreciate libraries like [bounce](https://github.com/hapijs/bounce/blob/master/API.md) (though I think this may catered to server-side) which ensure developer/system errors don't get missed when permitting some errors.

PR #138, opened 2024-06-26
:sparkles: Show embedded list in post card

foysalit:
Screenshot 2024-06-27 at 00 46 03

> Before

Screenshot 2024-06-27 at 00 39 40
 > After

### DIFF
--- a/components/common/posts/PostsFeed.tsx
+++ b/components/common/posts/PostsFeed.tsx
@@ -340,7 +340,7 @@ export function RecordEmbedView({
       rkey: `${rkey}`,
       collection: `lists`,
     })
-    console.log(embed.record)
+
     return (
       <div
         className={`flex gap-2 pb-2 ${leftPadding} flex-col border-2 border-gray-400 border-dashed my-2 rounded pt-2`}
--- a/components/common/posts/PostsFeed.tsx
+++ b/components/common/posts/PostsFeed.tsx
@@ -7,6 +7,8 @@ import {
   AppBskyEmbedRecordWithMedia,
   AppBskyFeedPost,
   AppBskyEmbedRecord,
+  AppBskyGraphList,
+  AppBskyGraphDefs,
 } from '@atproto/api'
 import Link from 'next/link'
 import {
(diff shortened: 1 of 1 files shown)


components/common/posts/PostsFeed.tsx
```diff
@@ -331,6 +333,67 @@ export function RecordEmbedView({
         </div>
       </div>
     )
+  } else if (AppBskyGraphDefs.isListView(embed.record)) {
+    const { did, rkey } = parseAtUri(embed.record.uri) || {}
+    const peekLink = buildBlueSkyAppUrl({
+      did: `${did}`,
+      rkey: `${rkey}`,
```
devinivy, line 340: Is this a way of casting `did` and `rkey` into a string?  If so, could it end-up casting them to something we don't intend like `'null'` or `'undefined'`?

components/common/posts/PostsFeed.tsx
```diff
@@ -331,6 +333,67 @@ export function RecordEmbedView({
...
+      >
+        <div className="flex flex-row gap-1">
+          <ProfileAvatar
+            profile={{
+              avatar: embed.record.avatar,
+              did: embed.record.creator.did,
+              handle: embed.record.name,
+            }}
+            className="w-6 h-6 rounded-full"
+          />
```
devinivy, line 356: Would it be worth renaming the `ProfileAvatar` component or extracting out a more generic component from it?  I notice that we're "tricking" it into thinking it's displaying profile information.

components/common/posts/PostsFeed.tsx
```diff
@@ -331,6 +333,67 @@ export function RecordEmbedView({
...
+          <p className="text-sm font-medium text-gray-900 dark:text-gray-200 pl-2">
+            <Link
+              href={`/repositories/${embed.record.uri.replace('at://', '')}`}
+              className="hover:underline"
+            >
+              {embed.record.name ? (
+                <>
+                  <span className="font-bold">{embed.record.name}</span>
+                </>
```
devinivy, line 365: We should be able to get rid of this fragment.

devinivy (approved):

components/common/posts/PostsFeed.tsx
foysalit: hmm.. idk, it's minimal lift but I also don't see avatars being shown for anything other than users/profiles so don't think it it matters a lot either way?

components/common/posts/PostsFeed.tsx
foysalit: record.uri is always present and parsing the uri should always return did and rkey for isViewBlocked AFAICT so this is just appeasing the ts linter.

foysalit:
I don't disagree strongly with any of the suggestions above so might end up making those changes in a future iteration but don't think any of those are blockers so going to ahead and merge.

PR #139, opened 2024-06-27
:bug: Fix permission fetching

foysalit:
Since next.js generates static pages on build, attempting to access `client.session` in any of the paths that may get rendered in the static build triggers an auth error. So, this PR refactors the permission checks into a helper function that handles said error.

### DIFF
--- a/lib/server-config.ts
+++ b/lib/server-config.ts
@@ -1,4 +1,5 @@
 import { ToolsOzoneServerGetConfig, ToolsOzoneTeamDefs } from '@atproto/api'
+import client from './client'
 
 export type ServerConfig = {
   pds?: string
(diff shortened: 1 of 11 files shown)


lib/server-config.ts
```diff
@@ -36,6 +38,19 @@ export const parseServerConfig = (
...
+export const checkPermission = (
+  permission: keyof ServerConfig['permissions'],
+) => {
+  try {
+    return client.session.serverConfig?.permissions[permission]
+  } catch (e) {
+    // Trying to access client.session while unauthenticated throws an error in which case,
+    // we can safely assume the user does not have the permission
+    return false
+  }
```
devinivy, line 55: This works, but could also rely on `client.authState` to avoid leaning on throwing:
```ts
if (client.authState === AuthState.LoggedOut) {
  return false
}
return client.session.serverConfig?.permissions[permission]
```

foysalit: hmm... I know these are directly linked but feels a bit odd just within this context to assume that logged in state guarantees existnence of session. I'll leave it as if that's ok?

PR #144, opened 2024-07-01
:sparkles: Generic RepoFinder component

foysalit:
Add a generic RepoFinder input field that searches by handle/did and shows profile details

### DIFF
--- a/components/repo/Finder.tsx
+++ b/components/repo/Finder.tsx
@@ -122,7 +122,7 @@ export function RepoFinder({
         ) : (
           items.map((item) => (
             <Combobox.Option
-              key={item.handle}
+              key={item.did}
               value={selectionType === 'did' ? item.did : item.handle}
               className={({ active }) =>
                 `cursor-pointer p-2 flex items-center space-x-3 ${
--- a/components/repo/Finder.tsx
+++ b/components/repo/Finder.tsx
@@ -1,0 +1,150 @@
+import { useState, useEffect } from 'react'
+import { Combobox } from '@headlessui/react'
+import client from '@/lib/client'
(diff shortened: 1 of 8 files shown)


components/repo/Finder.tsx
```diff
@@ -0,0 +1,150 @@
...
+          <div className="p-2">
+            <div className="font-semibold text-sm">No results found</div>
+            <div className="text-sm">
+              Queries must be partial handle or full DIDs
+            </div>
+          </div>
+        ) : (
+          items.map((item) => (
+            <Combobox.Option
+              key={item.handle}
```
devinivy, line 125: This works just fine, but `item.did` may be a more natural identifier to use.

components/repo/Finder.tsx
```diff
@@ -0,0 +1,150 @@
...
+    }
+  }, [query])
+
+  return (
+    <Combobox
+      value={selectedItem}
+      onChange={(item) => {
+        setSelectedItem(item)
+        onChange(item)
+      }}
```
devinivy, line 99: It is a little complicated to have both internal state and external state working at the same time, as it means consumers may be prone to get out of sync with the internal state of the component.  Would it be awkward to make `Finder` a standard controlled form component?

devinivy (approved):

components/repo/Finder.tsx
foysalit: surely doable but only issue is that in a lot of places, we use uncontrolled inputs and then utilize web native html form methods to get data. I like that pattern for the most part and only use controlled method when absolutely needed so I left the default behavior to match that.

will do a refactor if we end up seeing issues with this.

PR #147, opened 2024-07-05
:bug: Re-render cmd palette results when typeaheadResults are fetched

foysalit:


foysalit:
Going to merge this without review since Devin is out till monday. It's a small change in hook dependency.

PR #149, opened 2024-07-19
:bug: Fix removing duplicate labels on different cids

foysalit:


app/actions/ModActionPanel/QuickAction.tsx
```diff
@@ -336,26 +336,29 @@ function Form(
         const negatingLabelsByCid: Record<string, string[]> = {}
 
         if (recordInfo?.labels?.length && 'cid' in subjectInfo) {
+          // go through each label we intended to remove
           labels.negateLabelVals.forEach((label) => {
-            const existingLabelWithDifferentCid = recordInfo.labels?.find(
-              ({ val: originalLabel, cid, src }) => {
-                return (
-                  originalLabel === label &&
-                  cid !== subjectInfo.cid &&
```
devinivy, line 344: I notice that this check has disappeared in the new version.  I think it served to avoid double-removing the label from the current CID (once from `negatingLabelsByCid` and then again from handling `coreEvent.labels`).

devinivy: I wonder if another way to go about this which would be a smaller change is to replace:
```diff
- const existingLabelWithDifferentCid = recordInfo.labels?.find(
+ const existingLabelWithDifferentCid = recordInfo.labels?.filter(
```

And then handle the resulting changes.  I think this captures pretty concisely the idea here that there may be _multiple_ versions with the label that need to be removed rather than just one.  (Assuming that I am understanding the bug that you're fixing, which I may be partially misunderstanding!)

devinivy (approved):
Had a question, but otherwise looks good!

app/actions/ModActionPanel/QuickAction.tsx
foysalit: 1. double removing shouldn't happen because we are emptying `coreEvent.negateLabelVals` for each label being removed.
2. I looked at the filter approach but with that, we still have to go through all matched labels that are not of the current cid so the control flow ends up using similar loops and checks.

just for the peace of my mind, I also added a cypress test to assert this implementation works.

PR #150, opened 2024-07-21
:sparkles: Graceful fallback if record loading fails

foysalit:
This PR passes admin token when calling `getList` so that takendown records are returned. It also adds graceful fallback for any record loading failure so that we at least show the skeleton UI that lets moderators action the record being viewed.

### DIFF
--- a/app/repositories/[id]/[...record]/page-content.tsx
+++ b/app/repositories/[id]/[...record]/page-content.tsx
@@ -75,10 +75,11 @@ export default function RecordViewPageContent({
 
       const uri = createAtUri({ did, collection, rkey })
       const getRecord = async () => {
-        const { data: record } = await client.api.tools.ozone.moderation.getRecord(
-          { uri },
-          { headers: client.proxyHeaders() },
-        )
+        const { data: record } =
+          await client.api.tools.ozone.moderation.getRecord(
+            { uri },
+            { headers: client.proxyHeaders() },
+          )
         return record
       }
       const getThread = async () => {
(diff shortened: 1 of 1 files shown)


app/repositories/[id]/[...record]/page-content.tsx
```diff
@@ -103,17 +104,24 @@ export default function RecordViewPageContent({
...
         getRecord(),
         getListProfiles(),
         getThread(),
       ])
-      return { record, thread, profiles }
+      return {
+        record: record.status === 'fulfilled' ? record.value : undefined,
+        thread: thread.status === 'fulfilled' ? thread.value : undefined,
+        profiles: profiles.status === 'fulfilled' ? profiles.value : undefined,
+      }
```
Comment (devinivy) [inline app/repositories/[id]/[...record]/page-content.tsx:124]: By default I always avoid swallowing errors since it complicates debugging when an unexpected issue is going on.  Totally your call whether to handle that here (e.g. by logging the errors), but figured I'd mention it!

devinivy (approved):

app/repositories/[id]/[...record]/page-content.tsx
Comment (foysalit) [reply app/repositories/[id]/[...record]/page-content.tsx:124]: we were kinda swallowing errors along this path already and the idea is that surfacing these errors were not helpful at all and would manifest as "failed to load record" or something along that line. most of these fetches are only adding additional content to the primary view so IMO this does not pose any issues.

Comment (devinivy) [reply app/repositories/[id]/[...record]/page-content.tsx:124]: Just to clarify— I was imagining preserving the error for developers (not for users).  In my experience when an application has a habit of obscuring errors from devs it can become difficult to detect when an unanticipated issue is going on.

I recognize my style on this is maybe a little bit on the far end of the spectrum though 😆 With a few exceptional cases, I default to always ensure errors are either a. handled explicitly (e.g. ignoring specific anticipated error conditions) or b. logged.

PR #155, opened 2024-07-31
🚧 Add post embedded video player

foysalit:


components/common/video/player.tsx
```diff
@@ -0,0 +1,47 @@
...
+      }
+    }
+  }, [hls])
+
+  useEffect(() => {
+    if (ref.current) {
+      if (Hls.isSupported()) {
+        hls.loadSource(source)
+      } else {
+        // TODO: fallback
```
devinivy, line 30: Is a fallback still needed here?

devinivy (approved):

components/common/video/player.tsx
foysalit: yeah would be nicer to just show a fallback message on unsupported browsers. added that in just now.

devinivy (approved):

devinivy (approved):

PR #156, opened 2024-08-02
🐛 Fix new labeler config steps

foysalit:
This PR ensures that initial labeler set up steps can continue on without getting blocked by attempts to fetch server config.

### DIFF
--- a/app/actions/ModActionPanel/QuickAction.tsx
+++ b/app/actions/ModActionPanel/QuickAction.tsx
@@ -50,6 +50,7 @@ import { DM_DISABLE_TAG } from '@/lib/constants'
 import { MessageActorMeta } from '@/dms/MessageActorMeta'
 import { ModEventDetailsPopover } from '@/mod-event/DetailsPopover'
 import { LockClosedIcon } from '@heroicons/react/24/solid'
+import { checkPermission } from '@/lib/server-config'
 
 const FORM_ID = 'mod-action-panel'
 const useBreakpoint = createBreakpoint({ xs: 340, sm: 640 })
@@ -206,6 +207,7 @@ function Form(
   const deactivatedAt = getDeactivatedAt(
     repo ? { repo } : record ? { record } : {},
   )
+  const canManageChat = checkPermission('canManageChat')
 
   // navigate to next or prev report
   const navigateQueue = (delta: 1 | -1) => {
@@ -590,7 +592,7 @@ function Form(
                 />
               </FormLabel>
             )}
-            {isSubjectDid && (
+            {isSubjectDid && canManageChat && (
               <div className="mb-3">
                 <MessageActorMeta did={subject} />
               </div>
--- a/components/common/labels/List.tsx
+++ b/components/common/labels/List.tsx
@@ -97,7 +97,7 @@ export const ModerationLabel = ({
   const labelVal = toLabelVal(label, recordAuthorDid)
   const isSelfLabeled = isSelfLabel(labelVal)
   const labelDefFromService =
-    labelerServiceDef?.policies.definitionById[label.val]
+    labelerServiceDef?.policies.definitionById?.[label.val]
   const labelerProfile = labelerServiceDef?.creator
   const labelClassNames = getLabelChipClassNames({
     label,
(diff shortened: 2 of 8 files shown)


components/shell/ConfigurationFlow.tsx
```diff
@@ -241,6 +243,19 @@ function IdentityConfigurationFlow({
             placeholder="Confirmation Code"
             className="block w-full mt-2"
             value={token}
+            autoFocus
```
foysalit, line 246: These are unrelated UX improvements.

app/actions/ModActionPanel/QuickAction.tsx
```diff
@@ -206,6 +207,7 @@ function Form(
   const deactivatedAt = getDeactivatedAt(
     repo ? { repo } : record ? { record } : {},
   )
+  const canManageChat = checkPermission('canManageChat')
```
foysalit, line 210: Also unrelated but this hides the `show messaging data` block from UI for external labelers.

components/common/labels/List.tsx
```diff
@@ -97,7 +97,7 @@ export const ModerationLabel = ({
   const labelVal = toLabelVal(label, recordAuthorDid)
   const isSelfLabeled = isSelfLabel(labelVal)
   const labelDefFromService =
-    labelerServiceDef?.policies.definitionById[label.val]
+    labelerServiceDef?.policies.definitionById?.[label.val]
```
foysalit, line 100: This can crash a page if a plain label is added when the policy definition does not exist. So, this is more of a safety fallback.

components/shell/ConfigurationFlow.tsx
```diff
@@ -164,7 +164,9 @@ export function ConfigurationFlow() {
         key={config.updatedAt}
         session={session}
         onComplete={(skip) => {
-          client.reconfigure({ skipRecord: skip })
+          client
+            .reconfigure({ skipRecord: skip })
+            .then(() => client.refetchServerConfig())
```
devinivy, line 169: I wouldn't expect the server config to be updated from the labeler record configuration flow—is it possible we don't need it?

lib/client.ts
```diff
@@ -79,7 +79,12 @@ class ClientManager extends EventTarget {
       password,
     })
     const config = await this._getConfig()
-    const serverConfig = await this._getServerConfig(agent, config.did)
+    // If the labeler needs further configuration, let's generate a mock serverConfig
+    // and let the user move forward with configuration steps
+    const needsConfiguration = config.needs.key || config.needs.service
+    const serverConfig = await (needsConfiguration
+      ? parseServerConfig({})
+      : this._getServerConfig(agent, config.did))
```
devinivy, line 87: I'm not sure I understand how this part works.  Ideally IMO the server config and the labeler config could be managed independently from each other, and we wouldn't need to avoid loading the real server config.  But maybe there's an overlap that I'm overlooking.

foysalit: requests to the server config seems to fail until the labeler config is available. the error was something like `record not found` or something along that line and since it doesn't make a lot of sense to fetch this until the labeler config is available, delaying the fetch makes sense to me.

components/shell/ConfigurationFlow.tsx
foysalit: it doesn't update but this is the issue i was seeing https://github.com/bluesky-social/ozone/pull/156#discussion_r1705843465 so after the configuration is complete, we are fetching the server config for real here.

lib/client.ts
devinivy: (And if we're not sure whether we interpreted the "record not found" error properly, perhaps there's still an issue out there!)

devinivy: Ohhhhkay, I get what's going on now.  They are tangled together a little bit because we need to use service proxying to fetch the server config.  And server proxying is not going to work until the service is reflected in the labeler's DID document.  So we need to wait until `config.needs.service` is `false` before we can make authed requests to ozone.

I made some small tweaks in https://github.com/bluesky-social/ozone/pull/156/commits/fc7093cb460d98ba09518873a91d9788123fc919, let me know what you think!

devinivy (approved):
Sweet!  I tried this out, working well 👍

PR #161, opened 2024-08-13
:bug: Don't show external label in label selector

foysalit:
When taking label action, the label selector shows all labels that are currently applied on the subject and as a result, any label applied by an external labeler can be `unselected`. However, in reality, this doesn't do anything since one can not remove a label applied by an external labeler. 

To reduce confusion around this, this PR removes external labelers from label selector.

### DIFF
--- a/app/actions/ModActionPanel/QuickAction.tsx
+++ b/app/actions/ModActionPanel/QuickAction.tsx
@@ -670,8 +670,8 @@ function Form(
                       name="labels"
                       formId={FORM_ID}
                       defaultLabels={currentLabels.filter((label) => {
-                        const serviceDid = client.getServiceDid()
-                        const isExternalLabel = allLabels.find((l) => {
+                        const serviceDid = client.getServiceDid()?.split('#')[0]
+                        const isExternalLabel = allLabels.some((l) => {
                           return l.val === label && l.src !== serviceDid
                         })
                         return !isSelfLabel(label) && !isExternalLabel
--- a/app/actions/ModActionPanel/QuickAction.tsx
+++ b/app/actions/ModActionPanel/QuickAction.tsx
@@ -669,9 +669,13 @@ function Form(
                       id="labels"
                       name="labels"
                       formId={FORM_ID}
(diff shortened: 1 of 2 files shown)


cypress/fixtures/seed.json
```diff
@@ -37,7 +37,7 @@
                   "labels":[
                       {
                           "ver":1,
-                          "src":"did:plc:4wke3qtisoohr7tf6vfhe5r2",
```
foysalit, line 40: these had to be changed because we never used the src value for any functionality being tested, now that we're hitting that path in tests, we need to make sure the src matches with the test dataset.

app/actions/ModActionPanel/QuickAction.tsx
```diff
@@ -669,9 +669,13 @@ function Form(
                       id="labels"
                       name="labels"
                       formId={FORM_ID}
-                      defaultLabels={currentLabels.filter(
-                        (label) => !isSelfLabel(label),
-                      )}
+                      defaultLabels={currentLabels.filter((label) => {
+                        const serviceDid = client.getServiceDid()
```
devinivy, line 673: Is it possible that this "DID" contains the service id `#atproto_labeler`, yet `l.src` below will omit that?  Or does `getServiceDid()` return just the DID without the service id on the end?

app/actions/ModActionPanel/QuickAction.tsx
```diff
@@ -669,9 +669,13 @@ function Form(
...
+                        const isExternalLabel = allLabels.find((l) => {
+                          return l.val === label && l.src !== serviceDid
+                        })
```
devinivy, line 676: No biggie, but if we're trying to get a bool out (as indicated by the name `isExternalLabel`) then `some()` does that for us.
```suggestion
                        const isExternalLabel = allLabels.some((l) => {
                          return l.val === label && l.src !== serviceDid
                        })
```

devinivy (approved):

foysalit:
made the changes requested and merging but happy to make other changes if needed.

PR #162, opened 2024-08-18
:sparkles: Request data from appview without auth if author is blocking mod service

foysalit:
Use an unauthed appview agent for fetching author feed directly from appview if the request through ozone fails.

Screenshot 2024-10-15 at 11 25 45


components/common/feeds/AuthorFeed.tsx
```diff
@@ -22,15 +22,55 @@ const getAuthorFeed = async ({ id, pageParam, typeFilter, options }) => {
...
+      // without passing any auth headers to ensure that the block behavior is not triggered
+      // One downside of this is that the feed will not include takendown posts as it normally would
+      // if the request was going through mod service that has admin access on the appview
+      if (e instanceof AppBskyFeedGetAuthorFeed.BlockedByActorError) {
+        // There are usually 2 header values passed, atproto-proxy and atproto-accept-labelers
+        // we want to keep the atproto-accept-labelers header so that all the labels are included in the response
+        delete options.headers['atproto-proxy']
+        // @TODO: move the appview url to somewhere more centralized
+        const url = new URL(
+          'https://api.bsky.app/xrpc/app.bsky.feed.getAuthorFeed',
```
bnewbold, line 49: Might want to make the host configurable here? In some cases it might make sense to use `public.api.bsky.app`, or in the future folks might want to use alternative appviews. Or we might want to point at a specific appview instance (eg, in a specific datacenter).

bnewbold:
The context here is at least partially independent mod services which don't have admin access to the appview anyways. This seems roughly right to me, i'll defer to Devin for typescript code review.

components/common/feeds/AuthorFeed.tsx
devinivy: The ozone instances already are configured with an appview, and the UI has access to that config— I think it would make sense to use the same one that the ozone backend is configured with.

components/common/feeds/AuthorFeed.tsx
```diff
@@ -22,15 +22,55 @@ const getAuthorFeed = async ({ id, pageParam, typeFilter, options }) => {
...
+            return
+          }
+          url.searchParams.append(key, value)
+        })
+
+        const response = await fetch(url.toString(), {
+          headers: { ...options.headers },
+          body: null,
+          method: 'GET',
+        })
```
devinivy, line 63: I think we should still be able to use an `AtpAgent` to make the request.  There are several things the raw `fetch()` wont do for us, including response validation.

components/common/feeds/AuthorFeed.tsx
```diff
@@ -22,15 +22,55 @@ const getAuthorFeed = async ({ id, pageParam, typeFilter, options }) => {
...
+        data = await response.json()
```
devinivy, line 64: This `data` could be an error response if we didn't get a `200`.  I think using an `AtpAgent` pointed directly at the appview to make the request is probably the move.

components/shell/ConfigurationContext.tsx
```diff
@@ -63,6 +64,13 @@ export const ConfigurationProvider = ({
     isLoading: isServerConfigLoading,
   } = useServerConfigQuery(labelerAgent)
 
+  const appviewAgent = useMemo<Agent | null>(() => {
+    if (serverConfig?.appview) {
+      return new Agent(new CredentialSession(new URL(serverConfig.appview)))
+    }
+    return null
+  }, [serverConfig?.appview])
+
```
matthieusieben, line 73: The `ConfigurationProvider` logic is already pretty complex on its own. I thing that keeping this logic in its own hook (in another file even) would make things much easier to maintain.

Something like so (didn't try):

```ts
export function useAppviewAgent() {
  const { appview } = useConfigurationContext().serverConfig

  return useMemo<Agent | null>(() => {
    if (appview) {
      return new Agent(new CredentialSession(new URL(appview))))
    }
    return null
  }, [appview)])
}
```

Or, if you prefer a long lived agent instance (instead of one that gets created per component), use an `AppviewAgentProvider` (I don't think this is necessary though).

components/shell/ExternalLabelersContext.tsx
```diff
@@ -61,7 +61,11 @@ export const ExternalLabelersProvider = ({
   // Keep the labelers header up-to-date with the external labelers
   useEffect(() => {
     labelerAgent.configureLabelers([config.did, ...externalLabelers])
-  }, [labelerAgent, config.did, externalLabelers])
+    // If we have an appviewAgent, make sure it respects all configured labelers
+    if (appviewAgent) {
+      appviewAgent.configureLabelers([config.did, ...externalLabelers])
+    }
+  }, [labelerAgent, appviewAgent, config.did, externalLabelers])
```
matthieusieben, line 68: I would split this into two distinct effects. Indeed, with the current implementation, if `appviewAgent` is updated, but `labelerAgent` isn't, `labelerAgent` will be re configured anyways. This is very much a nit in this particular case, but a good practice in general (avoiding un-necessary side effects).

matthieusieben (changes requested):

components/common/feeds/AuthorFeed.tsx
```diff
@@ -18,10 +18,12 @@ export const useAuthorFeedQuery = ({
 }) => {
   const { data: repoData } = useRepoAndProfile({ id })
   const labelerAgent = useLabelerAgent()
+  const appviewAgent = useAppviewAgent()
 
   return useInfiniteQuery({
     queryKey: ['authorFeed', { id, query, typeFilter }],
     queryFn: async ({ pageParam }) => {
+      let isFromAppview = false
```
matthieusieben, line 26: From the looks of it (and for someone that is not familiar with how everything works here), the loop body could therorically return results from both the labeler AND the appview. Because that is the case, using a single `boolean` to store `isFromAppview` could be misleading.
Additionnaly, it seems that if the labeller errors on the first loop iteration, and more items are still required, the labeller will un-necessarily be tryed again during the following loop iterations.

components/shell/ConfigurationContext.tsx
matthieusieben: Also note: If you are not going to use session manageement here, just do:
```ts
new Agent({ service: appview })
```

foysalit: I thought it would be nicer to keep the definition of pds and appview agent in the same place but not opposed to splitting them up.
the `new Agent()` instantiation doesn't work without a `SessionManager` instance so I'm not sure how I'd do `new Agent(appview)`? but yeah, agreed that the instantiation should be simpler if I don't need session management, if the api allowed it.

components/common/feeds/AuthorFeed.tsx
foysalit: the only time when the loop may use both labelerAgent and appviewAgent is if someone blocks the labeler WHILE the loop is being executed which is super unlikely so I think we can get away with this for a bit. other approach may be to check for a block record first and then use the right agent which is an additional api req for every author feed load so I think the current approach is a tiny bit better though, kinda complex.

good catch on the extra failing call, cleaned it up.

components/repositories/AccountView.tsx
```diff
@@ -234,7 +234,7 @@ export function AccountView({
                     <Posts id={id} onReport={setReportUri} />
                   )}
                   {currentView === Views.Follows && (
-                    <Follows count={profile?.followersCount} id={id} />
+                    <Follows count={profile?.followsCount} id={id} />
```
foysalit, line 237: Unrelated bugfix

components/shell/ConfigurationContext.tsx
```diff
@@ -164,3 +164,14 @@ export function useServerConfig() {
 export function usePermission(name: PermissionName) {
   return useServerConfig().permissions[name]
 }
+
+export function useAppviewAgent() {
+  const { appview } = useConfigurationContext().serverConfig
+
+  return useMemo<Agent | null>(() => {
+    if (appview) {
+      return new Agent(new CredentialSession(new URL(appview)))
```
matthieusieben, line 173: Doesn't this work ?

```suggestion
      return new Agent(appview)
```

matthieusieben (approved):
Look good :-)

components/repositories/AccountView.tsx
matthieusieben: 👍

components/shell/ConfigurationContext.tsx
foysalit: unfortunately, no. tried to explain that here https://github.com/bluesky-social/ozone/pull/162#discussion_r1801822826

Screenshot 2024-10-16 at 20 27 50

PR #163, opened 2024-08-18
🚧 Set list viewer and editor

foysalit:
https://github.com/user-attachments/assets/fe04c62c-d8b6-4235-a8a0-0910b71d95db


devinivy (approved):

components/sets/utils.ts
```diff
@@ -0,0 +1,9 @@
+export const createSetPageLink = (queryParams: Record<string, string>) => {
+  const url = new URL(window.location.href.replace(window.location.search, ''))
```
devinivy, line 2: Does the base URL need to be dynamic like this?  Could it look something like:
```ts
  const searchParams = new URLSearchParams()
  Object.entries({ tab: 'sets', ...queryParams }).forEach(([key, value]) => {
    searchParams.set(key, value)
  })
  return `/base/page/path?${searchParams.toString()}`
```

devinivy (approved):

components/sets/utils.ts
foysalit: yeah there's at least one place where we want to navigate to a nested view while maintaining the parent view so this is the safest way of making sure that we don't lose parent view because of param change.

PR #170, opened 2024-09-02
:bug: Allow Only/Exclude/Indifferent filters for appeal status

foysalit:
Depends on https://github.com/bluesky-social/atproto/pull/2777

https://github.com/user-attachments/assets/e36fc543-99db-4378-854b-a62d3db043bf

### DIFF
--- a/app/reports/page-content.tsx
+++ b/app/reports/page-content.tsx
@@ -63,7 +63,7 @@ const buildPageTitle = ({
   currentTab: string
   takendown: boolean
   includeMuted: boolean
-  appealed: boolean
+  appealed: string | null
 }) => {
   const titleFromTab =
     currentTab === 'all'
@@ -79,8 +79,12 @@ const buildPageTitle = ({
     additionalFragments.push('Include Muted')
   }
 
-  if (appealed) {
-    additionalFragments.push('Appealed')
+  if (appealed === null) {
+    if (appealed === 'true') {
+      additionalFragments.push('Only Appeals')
+    } else {
+      additionalFragments.push('No Appeals')
+    }
   }
 
   const additionalTitle = additionalFragments.length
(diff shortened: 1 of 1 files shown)


app/reports/page-content.tsx
```diff
@@ -79,8 +79,12 @@ const buildPageTitle = ({
...
 
-  if (appealed) {
-    additionalFragments.push('Appealed')
+  if (appealed !== null) {
+    if (appealed === 'true') {
+      additionalFragments.push('Only Appeals')
+    } else {
+      additionalFragments.push('No Appeals')
+    }
   }
```
devinivy, line 88: Just lining the logic up with similar logic I see below.
```suggestion
  if (appealed === 'true') {
    additionalFragments.push('Only Appeals')
  } else if (appealed === 'false') {
    additionalFragments.push('No Appeals')
  }
```

app/reports/page-content.tsx
```diff
@@ -232,7 +250,9 @@ export const ReportsPageContent = () => {
         }
 
         if (appealed) {
-          queryParams.appealed = appealed
+          // If not specifically set to true but there is a value, we can default to false
+          // No value will pass undefined which will be ignored
+          queryParams.appealed = appealed === 'true'
         }
```
devinivy, line 256: I wonder if an unknown value would be better suited to be interpreted as "unset" rather than `false`, and some of the logic above seems to handle it that way.

devinivy (approved):

app/reports/page-content.tsx
devinivy: E.g. something like:
```js
        if (appealed === 'true') {
          queryParams.appealed = true
        } else if (appealed === 'false') { 
          queryParams.appealed = false
        }
```

foysalit: hmm... I think I see what you mean but params.get returns null and setting it to an "unknown" value feels odd and could only be needed if we ever give this more meaning in which case we'd need another touchpoint here anyways. so I'll leave it as is for now but happy to discuss more if you feel strongly.

devinivy: The main thing I was noticing is that in some logic (mostly the display logic) an unknown value was treated as `false` and in other logic it was treated as unset.  So it looked like a consistency issue, e.g. the UI might display something different than what the server was receiving.

devinivy: (My inclination is it ignore unknown data and keep the default behavior rather than interpret it to a specific value.)

devinivy: Trying to illustrate better!  I expect the current behavior is like this:
 - `?appealed=`
   - UI shows "Appeals"
   - 👍 Server doesn't filter based on appeal
 - `?appealed=true`
   - UI shows "Only Appeals"
   - 👍 Server services "only appeals"
 - `?appealed=false`
   - UI shows "No Appeals"
   - 👍 Server services "no appeals"
 - `?appealed=unknown`
   - UI shows "Appeals"
   - ⚠️ Server services "no appeals"

foysalit: makes sense. will get it wired up in the next PR! thanks

PR #176, opened 2024-09-04
:sparkles: Add can't send email warning for users that can not receiv…

foysalit:
Screenshot 2024-09-04 at 12 33 29


bnewbold (approved):
UX LGTM

components/email/useEmailRecipientStatus.tsx
```diff
@@ -0,0 +1,33 @@
...
+        return false
+      }
+
+      const pdsEndpoint = response.services['atproto_pds']?.endpoint
+      if (!pdsEndpoint) {
+        return false
+      }
+
+      return DOMAINS_ALLOWING_EMAIL_COMMUNICATION.some((domain) => {
+        return pdsEndpoint.endsWith(domain)
```
devinivy, line 22: We probably want the domain to end with `.bsky.network` rather than `bsky.network`.

lib/constants.ts
```diff
@@ -40,3 +40,7 @@ export const YOUNG_ACCOUNT_MARKER_THRESHOLD_IN_DAYS = process.env
   .NEXT_PUBLIC_YOUNG_ACCOUNT_MARKER_THRESHOLD_IN_DAYS
   ? parseInt(process.env.NEXT_PUBLIC_YOUNG_ACCOUNT_MARKER_THRESHOLD_IN_DAYS)
   : 30
+
+export const DOMAINS_ALLOWING_EMAIL_COMMUNICATION = (
+  process.env.NEXT_PUBLIC_DOMAINS_ALLOWING_EMAIL_COMMUNICATION || 'bsky.network'
```
devinivy, line 45: Perhaps we should default to empty, and configure this specially on our deployments— otherwise this will be the configuration in the distro.

devinivy (approved):
Left a couple notes, but looking good 👍

components/email/useEmailRecipientStatus.tsx
foysalit: makes sense. but since this is env var, we can just configure it that way in the var, right? maybe the name needs an adjustment so that it's not DOMAINS_ALLOWING_EMAIL_COMMUNICATION and instead DOMAINS_SUFFIXES_ALLOWING_EMAIL_COMMUNICATION?

foysalit: @devinivy do lmk if you have thoughts on this, I'll merge for now and happy to address any different outcome from this as a fast follow.

devinivy: Yup we can just configure it that way 👍

PR #178, opened 2024-09-05
Workspace improvements

foysalit:
This PR adds a bunch of new improvements to workspace feature

1. Take action in batches. If there are too many actions to be taken, we will now batch 50 actions at a time and then add a cooling period after each batch before firing the next one.
2. Add all search results from repositories page to workspace. This allows mods to search for a user with a keyword and all matching users to workspace.
3. Add all followers of a user to workspace.
4 Add all follows of a user to workspace.

### DIFF
--- a/components/graph/Followers.tsx
+++ b/components/graph/Followers.tsx
@@ -1,0 +1,112 @@
+import { ActionButton } from '@/common/buttons'
+import { LoadMoreButton } from '@/common/LoadMoreButton'
+import { ConfirmationModal } from '@/common/modals/confirmation'
+import { AccountsGrid } from '@/repositories/AccountView'
+import { useLabelerAgent } from '@/shell/ConfigurationContext'
+import { useWorkspaceAddItemsMutation } from '@/workspace/hooks'
+import { useInfiniteQuery } from '@tanstack/react-query'
+import { useState } from 'react'
+import { toast } from 'react-toastify'
+
+export function Followers({ id, count }: { id: string; count?: number }) {
+  const [isConfirmationOpen, setIsConfirmationOpen] = useState(false)
+  const labelerAgent = useLabelerAgent()
+  const [isAdding, setIsAdding] = useState(false)
+  const { mutateAsync: addItemsToWorkspace } = useWorkspaceAddItemsMutation()
+  const { error, isLoading, data, fetchNextPage, hasNextPage } =
+    useInfiniteQuery({
+      queryKey: ['followers', { id }],
+      queryFn: async ({ pageParam }) => {
+        const { data } = await labelerAgent.api.app.bsky.graph.getFollowers({
+          actor: id,
+          cursor: pageParam,
+        })
+        return data
+      },
+      getNextPageParam: (lastPage) => lastPage.cursor,
+    })
+
+  const followers = data?.pages.flatMap((page) => page.followers) ?? []
+
+  const confirmAddToWorkspace = async () => {
+    // add items that are already loaded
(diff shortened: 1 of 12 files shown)


components/graph/Followers.tsx
```diff
@@ -0,0 +1,112 @@
...
+export function Followers({ id, count }: { id: string; count?: number }) {
+  const [isConfirmationOpen, setIsConfirmationOpen] = useState(false)
+  const labelerAgent = useLabelerAgent()
+  const [isAdding, setIsAdding] = useState(false)
+  const { mutateAsync: addItemsToWorkspace } = useWorkspaceAddItemsMutation()
+  const { error, isLoading, data, fetchNextPage, hasNextPage } =
+    useInfiniteQuery({
+      queryKey: ['followers', { id }],
+      queryFn: async ({ pageParam }) => {
+        const { data } = await labelerAgent.api.app.bsky.graph.getFollowers({
```
matthieusieben, line 20: nit: `.api.` no longer required (deprecated)

```suggestion
        const { data } = await labelerAgent.app.bsky.graph.getFollowers({
```

components/workspace/hooks.tsx
```diff
@@ -25,15 +26,34 @@ export const useWorkspaceList = () => {
...
+        )}, there are ${pluralize(
+          allItems.length,
+          'subject',
+        )} in your workspace now.`
+
+        if (!toastId.current) {
+          toastId.current = toast.success(message)
+        } else {
+          // Only when pass the onClose here because if we pass it during instantiation, the update essentially triggers onClose
+          // causing the toast to close indefinitely
```
matthieusieben, line 48: Doesn't updating twice also cause this behavior ?


Would keeping another reference help ?
```tsx
const toastNr = useRef<number>(0)
const toastId = useRef<number | string | null>(null)

// ...

onSuccess: () => {
  // ...

  const currentNr = ++toastNr.current
  const onClose = () => {
    // Only clear the toastId ref when the latest toast is closed
    if (toastNr.current === currentNr) toastId.current = null
  }

  // use onClose both for init and update
  // ...
},
```

components/workspace/Panel.tsx
```diff
@@ -123,6 +123,19 @@ export function WorkspacePanel(props: PropsOf<typeof ActionPanel>) {
...
+
+      // If there are any item that failed to action, we want to keep them checked so users know which ones to retry
+      if (results.failed.length) {
+        document
+          .querySelectorAll<HTMLInputElement>(
+            'input[type="checkbox"][name="workspaceItem"]',
+          )
+          .forEach((checkbox) => {
+            if (results.failed.includes(checkbox.value)) {
+              checkbox.checked = true
```
matthieusieben, line 135: If I'm not mistaken, this could cause the react component's internal state not to be in sync with the actual checkbox state, leading to unexpected behavior. Indeed, setting `checkbox.checked` won't trigger the change listener that the react component typically listens for in order to keep the internal state in sync.

matthieusieben (commented):
Looks good.

components/workspace/Panel.tsx
foysalit: good catch! the internal state is only kept for a UX detail which let's us do gmail style shift+click checkbox selection. which isn't that important to the functionalities. thanks for pointing this out, we're now triggering the mousedown event manually along with `checked` value update

components/workspace/hooks.tsx
foysalit: thanks ended up using the isActive check.

matthieusieben (approved):

PR #184, opened 2024-09-11
:sparkles: Export accounts from workspace to csv and add items from drag-n-drop csv

foysalit:
For record keeping, sharing data for further research and fill legal requests exporting data of a group of users is often done manually by mods. This is aimed to make that process much easier.

Also, allow drag-n-dropping a csv file and filling in workspace items from the csv content.

Screenshot 2024-12-06 at 23 33 32

> Export configuration

Screenshot 2024-12-07 at 01 02 00

> Workspace import file button small mode

Screenshot 2024-12-07 at 01 02 12

> Workspace import file button full mode

Screenshot 2024-12-07 at 01 02 27

> Workspace file import error


components/workspace/hooks.tsx
```diff
@@ -95,6 +98,77 @@ export const useWorkspaceEmptyMutation = () => {
...
+              ? (repo.relatedRecords[0] as AppBskyActorProfile.Record)
+              : null
+
+            const line: string[] = [
+              repo.did,
+              repo.handle,
+              repo.email || 'Unknown',
+              `${repo.ip || 'Unknown'}`,
+              `${profile?.displayName || 'Unknown'}`,
+              repo.labels?.map(({ val }) => val).join(', ') || 'None',
```
devinivy, line 159: I almost wonder if an empty value could be more useful than a default value here.  I suppose it depends how the CSV will be used.

components/workspace/hooks.tsx
```diff
@@ -95,6 +98,77 @@ export const useWorkspaceEmptyMutation = () => {
...
+              buildBlueSkyAppUrl({ did: repo.did }),
+            ]
+            return line.join(',')
```
devinivy, line 162: Values containing commas or newlines will end up breaking the CSV format.  If we add some escaping we should be able to address it, though:
```ts
const escapeCSVValue = (value: string) => {
  if (value.contains(',') || value.contains('"') || value.contains('\r') || value.contains('\n')) {
    return `"${value.replaceAll('"', '""')}"`
  }
  return value
}

// ...

  return line.map(escapeCSVValue).join(',')
```

components/workspace/hooks.tsx
foysalit: yeah I think our current use case is for mod folks to get these exports and review manually in which case, I feel like the explicit default values are more helpful.

components/workspace/hooks.tsx
foysalit: yh except for labels/tags list, we don't have anything that contain those characters though but will add to make it more future proof.

components/workspace/Panel.tsx
```diff
@@ -272,79 +276,126 @@ export function WorkspacePanel(props: PropsOf<typeof ActionPanel>) {
...
+                    className="h-10 w-10 text-green-300 align-text-bottom mx-auto mb-4"
+                  />
+                  <p className="pb-4 text-center text-gray-400 dark:text-gray-50">
+                    Workspace is empty.
+                  </p>
+                  <WorkspaceItemCreator
+                    onFileUploadClick={() => {
+                      dropzoneRef.current?.open()
+                    }}
+                  />
```
matthieusieben, line 318: This uses two different patterns:
- Data binding through props
- Data binding through (some sort) of context
Sadly I don't see an easy way to do "clean" this. If you do, I'd change this, otherwise let's leave it.

lib/csv.ts
```diff
@@ -0,0 +1,136 @@
...
+  headers: string[]
+  filename?: string
+  lineDelimiter?: string
+}) {
+  return {
+    filename: (filename || Date.now().toString()) + '.csv',
+    headerRow: headers.join(',') + lineDelimiter, // make your own csv head
+    body: lines.join(lineDelimiter),
+  }
+}
```
matthieusieben, line 33: The return value does not match the returned type. Is there a particular reason for that ?

The data structure to represent the CSV here is a mix between encoded data (a multi line string of comma separated values, with the firt line being a header) and "parsed" data (internal data structure), Looks like we could benefit from defining and using a type interface (or even a class ?) for internal representation of CSVs:

```ts
type CsvContent = {
  headers: string[],
  rows: [][]
}
```

lib/csv.ts
```diff
@@ -0,0 +1,136 @@
...
+  const rows = data.split('\n').map((row) => row.trim())
+  const [header, ...content] = rows
+
+  if (!header) return []
+
+  // In case header names are quoted, we want to exclude those quotes before check
+  const headers = header.split(',').map((col) => {
+    const trimmed = col.trim()
+    return trimmed.startsWith('"') && trimmed.endsWith('"')
+      ? trimmed.slice(1, -1)
```
matthieusieben, line 106: Does this logic accounts for cells containing `"` or `,` ?

lib/csv.ts
```diff
@@ -0,0 +1,136 @@
...
+
+  return content
+    .map((row) => {
+      const columns = row.split(',').map((col) => {
+        const trimmed = col.trim()
+        return trimmed.startsWith('"') && trimmed.endsWith('"')
+          ? trimmed.slice(1, -1)
+          : trimmed
+      })
+      return columns[didIndex] || columns[uriIndex]
```
matthieusieben, line 122: Looks like the row parsing and header parsing contain a lot of duplication

matthieusieben (approved):
Looks good. Some room for refactor, but nothing that needs to be done now.

lib/csv.ts
foysalit: cleaned em up thanks!

lib/csv.ts
foysalit: for parsing content we just want to extract did or at uri and neither of them can have those characters in the middle so this is really just to make sure that we can handle quoted csvs.

lib/csv.ts
```diff
@@ -0,0 +1,136 @@
...
+          reject(new Error(`Invalid CSV file: ${file.name}`))
+        }
+      }
+      reader.onerror = () => reject(new Error(`Failed to read ${file.name}`))
+      reader.readAsText(file)
+    } else {
+      reject(new Error(`Unsupported file type: ${file.name}`))
+    }
+  })
+}
```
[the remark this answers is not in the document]
foysalit: thanks, cleaned it up

lib/csv.ts
foysalit: > The return value does not match the returned type

not sure if I follow? the return type is expected to be `CsvContent` which i left implicit I guess. agreed we can tidy this up a bit more to make a more strict definition but for now the use-case is super low-stake so wanted to keep this flexible/open.

foysalit:
Jammed in one more feature here to allow profile bio/record content search in workspace filter panel.

PR #185, opened 2024-09-11
:sparkles: Recursively load queue pages if current page is empty

foysalit:
This avoids showing empty queue or hitting `Load More` and not getting new items when multiple queues are active.

### DIFF
--- a/app/reports/page-content.tsx
+++ b/app/reports/page-content.tsx
@@ -367,6 +367,7 @@ const getQueueItems = async (
   labelerAgent: Agent,
   queryParams: ToolsOzoneModerationQueryStatuses.QueryParams,
   queueName: string | null,
+  attempt = 0,
 ) => {
   const pageSize = 50
   const { data } = await labelerAgent.tools.ozone.moderation.queryStatuses({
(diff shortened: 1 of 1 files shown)


devinivy (approved):

app/reports/page-content.tsx
```diff
@@ -356,29 +357,50 @@ function useModerationQueueQuery() {
...
+        const queueDeciderCharCode =
+          `${subjectDid}`.split(':').pop()?.charCodeAt(0) || 0
+        return queueDeciderCharCode % queueDivider === queueIndex
+      })
+    : data.subjectStatuses
+
+  // This is a recursive call to get items in queue if the current page
+  // gives us less than full page size and there are more items to fetch
+  if (statusesInQueue.length === 0 && data.cursor) {
+    return getQueueItems(
```
devinivy, line 395: Only think I am a little unsure about here is that it would be easy to create a bug on the server side that causes this to spin and cause tons of requests.  The fix, I think, would be to pass a counter as a last arg to `getQueueItems()` that ensures we don't fetching more than maybe 10 pages this way.

devinivy (approved):

PR #187, opened 2024-09-11
:sparkles: Show subtitles in videos, if there are any

foysalit:
Screenshot 2024-09-11 at 20 21 50

> Subtitle option

Screenshot 2024-09-11 at 20 21 45

> Lang specific subtitles

### DIFF
--- a/components/common/posts/PostsFeed.tsx
+++ b/components/common/posts/PostsFeed.tsx
@@ -233,12 +233,14 @@ export function PostEmbeds({ item }: { item: AppBskyFeedDefs.FeedViewPost }) {
   )
 
   if (AppBskyEmbedVideo.isView(embed)) {
+    const captions = item.post.record?.['embed']?.['captions']
     return (
       <div className="flex gap-2 pb-2 pl-4" aria-label={embed.alt}>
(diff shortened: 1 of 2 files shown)


components/common/posts/PostsFeed.tsx
```diff
@@ -233,12 +233,14 @@ export function PostEmbeds({ item }: { item: AppBskyFeedDefs.FeedViewPost }) {
   )
 
   if (AppBskyEmbedVideo.isView(embed)) {
+    const captions = item.post.record?.['embed']?.['captions']
```
devinivy, line 236: Where the captions live in the record depends on whether it's a video embed, or a record-with-media embed.  Would recommend using the lexicon type guards to figure out where to pull it from (will also make it less likely that you get an invalid `captions` value that causes a UI crash).

devinivy (approved):

components/common/posts/PostsFeed.tsx
foysalit: ok I'm gonna merge this for now since this works. for cases that I'm missing here, will circle back and plug those up tomorrow.

PR #188, opened 2024-09-11
:sparkles: Add embed type filter for moderation queue

foysalit:
This allows moderators to look at reported content based on their embedded content types.

**Note**: It does not work with language filter since it uses tags for filtering and a backend limitation of how tag filter works as an OR condition instead of AND when there are multiple tags.

Screenshot 2024-09-11 at 20 47 46

> Light mode

Screenshot 2024-09-11 at 20 47 33

> Dark mode


devinivy (approved):
Ace!

PR #190, opened 2024-09-12
:bug: Uncheck ack all option by default when taking down

foysalit:


foysalit:
Merging without review to avoid mods accidentally acking a bunch of stuff without realizing this is checked.

PR #194, opened 2024-09-20
:lipstick: Make graphic media less visible

foysalit:
This PR adds a localstorage based preference that allows users to choose 3 possible filters on media content with certain labels. Previously, we were only applying a blur filter for all these labels and now, along with blur, media can be black and white and translucent, according to user's preference. 

**By Default we will apply all 3 filters for those labels**

Screenshot 2024-09-20 at 23 21 46

> Dark mode

Screenshot 2024-09-20 at 23 21 38

> Light mode

Additionally, this PR adds modal image viewer for profile avatar and profile banner image.

### DIFF
--- a/components/config/LocalPreferences.tsx
+++ b/components/config/LocalPreferences.tsx
@@ -1,0 +1,91 @@
+import { ActionButton } from '@/common/buttons'
+import { Card } from '@/common/Card'
+import { Checkbox } from '@/common/forms'
+import {
+  buildGraphicPreferenceKeyForLabel,
+  GraphicMediaFilter,
+  LabelChip,
+  labelsRequiringBlur,
+} from '@/common/labels'
+import { useGraphicMediaPreferences } from './useLocalPreferences'
+import { toast } from 'react-toastify'
+
+const GraphicMediaPreferenceSelectorForLabel = ({
+  label,
+}: {
+  label: string
+}) => {
+  const { getPreference } = useGraphicMediaPreferences()
+
+  return (
+    <div className="my-2">
+      <LabelChip className="ml-0 mb-2">{label}</LabelChip>
+      {(['blur', 'grayscale', 'translucent'] as GraphicMediaFilter[]).map(
+        (filter) => {
+          const key = buildGraphicPreferenceKeyForLabel(label, filter)
+          return (
+            <Checkbox
+              defaultChecked={getPreference(key)}
+              label={filter}
+              className="capitalize"
+              name={key}
+              key={key}
+              value="on"
+            />
+          )
+        },
+      )}
+    </div>
+  )
+}
+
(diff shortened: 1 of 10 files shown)


devinivy (approved):

components/config/LocalPreferences.tsx
```diff
@@ -0,0 +1,91 @@
...
+  label,
+}: {
+  label: string
+}) => {
+  const { getPreference } = useGraphicMediaPreferences()
+
+  return (
+    <div className="my-2">
+      <LabelChip className="ml-0 mb-2">{label}</LabelChip>
+      {(['blur', 'grayscale', 'translucent'] as GraphicMediaFilter[]).map(
```
devinivy, line 23: This list is used a few places and cast into `GraphicMediaFilter[]`.  Would it make sense to export it somewhere as a const with the proper type?

components/config/useLocalPreferences.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  labelsRequiringMediaFilter.forEach((label) => {
+    ;(['blur', 'grayscale', 'translucent'] as GraphicMediaFilter[]).forEach(
+      (filter) => {
+        initialValue.graphicMediaPrefs[
+          buildGraphicPreferenceKeyForLabel(label, filter)
+        ] = true
+      },
+    )
+  })
```
devinivy, line 26: I find this a bit difficult to read!  Would be sweet if we could get it to be something like:
```ts
labelsRequiringMediaFilter.forEach((label) => {
  GRAPHIC_MEDIA_FILTERS.forEach((filter) => {
    const key = buildGraphicPreferenceKeyForLabel(label, filter)
    initialValue.graphicMediaPrefs[key] = true
  })
})
```

components/config/useLocalPreferences.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  const [localPreferences, setLocalPreferences] = useLocalStorage(
+    'ozoneLocalPreferences',
```
devinivy, line 29: Would there be any value in clearing these on logout, or keeping separate keys per user?

devinivy (approved):

components/config/LocalPreferences.tsx
foysalit: totally, was just being lazy :(

components/config/useLocalPreferences.tsx
foysalit: I think I've avoided that until now since I see these configs as individual/device level preference, not account level. if the same user is accessing ozone using different accounts, I feel like they would want these configs across accounts as long as it's the same device. but yeah, I don't think the other approach to tie it to user level makes sense in some contexts.

PR #199, opened 2024-09-25
:bug: Open only action panel when subject is opened from workspace

foysalit:


components/reports/SubjectOverview.tsx
```diff
@@ -38,15 +38,20 @@ const CollectionLink = ({
...
   repoUrl: string
+  omitQueryParams?: string[]
 }) => {
   const pathname = usePathname()
   const searchParams = useSearchParams()
   const newUrl = new URLSearchParams(searchParams)
   newUrl.set('quickOpen', uri)
+  if (omitQueryParams?.length) {
+    omitQueryParams.forEach((param) => newUrl.delete(param))
+  }
```
matthieusieben, line 54: nit: could make this slightly shorter

```suggestion
omitQueryParams?.forEach((param) => newUrl.delete(param))
```

matthieusieben (commented):
looks good !

PR #206, opened 2024-09-30
:sparkles: Fetch workspace data in batches and add advanced filtering

foysalit:
Depends on https://github.com/bluesky-social/atproto/pull/2836

Use the newly available getRepos and getRecords endpoints to bulk fetch data for workspace items. This also allows us to provide additional context to workspace items without expanding into full view mode per item.

Screenshot 2024-09-30 at 17 09 32

This PR also adds a way to filter items from the workspace based on a combination of options among accounts and other content type which allows mods to select/unselect items from workspace in batches.

Screenshot 2024-10-04 at 20 46 33

### DIFF
--- a/components/common/labels/List.tsx
+++ b/components/common/labels/List.tsx
@@ -106,7 +106,7 @@ export const ModerationLabel = ({
   })
 
   return (
-    <Popover className="relative">
+    <Popover className="relative z-20">
       {({ open }) => (
         <>
           <Popover.Button className="ring-none">
--- a/components/workspace/Panel.tsx
+++ b/components/workspace/Panel.tsx
@@ -21,6 +21,7 @@ import { WorkspacePanelActions } from './PanelActions'
 import { WORKSPACE_FORM_ID } from './constants'
 import { WorkspacePanelActionForm } from './PanelActionForm'
 import { useActionSubjects } from '@/mod-event/helpers/emitEvent'
+import { useWorkspaceListData } from './useWorkspaceListData'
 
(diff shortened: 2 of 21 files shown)


components/common/labels/List.tsx
```diff
@@ -106,7 +106,7 @@ export const ModerationLabel = ({
   })
 
   return (
-    <Popover className="relative">
+    <Popover className="relative z-20">
```
foysalit, line 109: this known z-index value helps us overlap other content on labels.

components/workspace/Panel.tsx
```diff
@@ -35,23 +36,13 @@ export function WorkspacePanel(props: PropsOf<typeof ActionPanel>) {
...
-  }
-
   const handleRemoveSelected = () => {
     const selectedItems = Array.from(
       formRef.current?.querySelectorAll<HTMLInputElement>(
         'input[type="checkbox"][name="workspaceItem"]:checked',
       ) || [],
     ).map((checkbox) => checkbox.value)
-    removeItemsMutation.mutate(selectedItems)
+    removeItemsMutation.mutate(selectedItems as string[])
```
devinivy, line 45: Are the checkbox values always strings, never e.g. `undefined`?

devinivy (approved):
Nice 👍

components/workspace/Panel.tsx
foysalit: yep, these checkboxes always carry the value of the did/uri

PR #207, opened 2024-10-01
:sparkles: Add search page

foysalit:
This PR adds a dedicated page to show top and latest posts for a search keyword which surfaces the same content as the app. This helps moderators to quickly find and proactively action content without having to go through the app.

Screenshot 2024-10-01 at 20 10 14

### DIFF
--- a/components/search-content/useContentSearch.tsx
+++ b/components/search-content/useContentSearch.tsx
@@ -1,0 +1,38 @@
+import { useLabelerAgent } from '@/shell/ConfigurationContext'
+import { useWorkspaceAddItemsMutation } from '@/workspace/hooks'
+import { useInfiniteQuery } from '@tanstack/react-query'
+
+export const useContentSearch = ({
+  term,
+  section,
+}: {
+  term: string
+  section: string
+}) => {
+  const labelerAgent = useLabelerAgent()
+
+  const { mutate: addToWorkspace } = useWorkspaceAddItemsMutation()
+
+  const searchInfo = useInfiniteQuery({
+    queryKey: ['searchContent', { term, section }],
+    enabled: !!term,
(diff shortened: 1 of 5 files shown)


components/search-content/useContentSearch.tsx
```diff
@@ -0,0 +1,38 @@
...
+      const { data } = await labelerAgent.app.bsky.feed.searchPosts({
+        q: term,
+        limit: 30,
+        sort: section,
+        cursor: pageParam,
+      })
+      return data
+    },
+    getNextPageParam: (lastPage) => lastPage.cursor,
+  })
```
matthieusieben, line 29: I fear that this query could have two effects:
1) a lot of entries end up being cached, significantly increasing the memory usage
2) there isn't any mechanism to clear that cache so the user could end-up seeing stale results when trying to search for newer posts.

Is there a way to invalidate all `searchContent` cache entries, for example when the component gets destroyed ?

Alternatively, maybe a mutation would be more appropriate here ?

matthieusieben (commented):
Looks pretty straight forward. I do have a small concern about caching of search results.

components/search-content/useContentSearch.tsx
foysalit: good point but I think this is actually what we want. when investigating search results, I don't think a 5m delay in seeing realtime content is a big issue for mods and in fact, we wouldn't want the results to change while they are scanning a specific page for a keyword so overall, the caching seems to be a good idea.
keep in mind that at any given moment, a mod will probably scan for only a few keywords so keyword specific caches can be kept around without overwhelming memory usage IMO.

matthieusieben (approved):
If the caching ends-up being problematic, we can always adapt later ;-)

PR #211, opened 2024-10-03
💄 Replace lang tags with flag emoji

foysalit:
This PR replaces lang tags with the associated flag emoji, if there is one.

Screenshot 2024-10-03 at 21 40 00

> in quick action panel

Screenshot 2024-10-03 at 21 40 29

> in workspace

### DIFF
--- a/components/workspace/List.tsx
+++ b/components/workspace/List.tsx
@@ -174,7 +174,7 @@ const ListItem = <ItemType extends string>({
   // Derive language tag from record value if there isn't any tag in moderation.subjectStatus
   // which happens when a post has not been in the moderation system yet so we never tagged its language
   const langTagsFromRecord =
-    !isRepo && itemData ? getLangTagFromRecordValue(itemData) : []
+    isRecord && itemData ? getLangTagFromRecordValue(itemData) : []
 
   const subjectStatus = getSubjectStatusFromItemData(itemData)
   let repoHandle = getRepoHandleFromItemData(itemData)


lib/locale/languages.ts
```diff
@@ -2,545 +2,2612 @@ interface Language {
...
-  { code3: 'alb', code2: 'sq', name: 'Albanian' },
-  { code3: 'ale', code2: '', name: 'Aleut' },
-  { code3: 'alg', code2: '', name: 'Algonquian languages' },
-  { code3: 'alt', code2: '', name: 'Southern Altai' },
-  { code3: 'amh', code2: 'am', name: 'Amharic' },
-  { code3: 'ang', code2: '', name: 'English, Old (ca.450-1100)' },
-  { code3: 'anp ', code2: 'Angika', name: 'Angika' },
-  { code3: 'apa', code2: '', name: 'Apache languages' },
-  { code3: 'ara', code2: 'ar', name: 'Arabic' },
+  {
```
matthieusieben, line 9: nit: Wouldn't this be more readable using 1 line per item?

matthieusieben: Also, would 'this be more memory efficient by omitting `code2` if empty?

matthieusieben (approved):
looks good. Any idea why the build failed ?

https://github.com/bluesky-social/ozone/actions/runs/11187349902/job/31104207001?pr=211

lib/locale/languages.ts
foysalit: prettier automatically formatted these but also, once compiled, I don't think it matters much, right?
cod3 is used for a different feature I think, can't recall exactly so we kinda need every item here.

foysalit:
> looks good. Any idea why the build failed ?
> 
> https://github.com/bluesky-social/ozone/actions/runs/11187349902/job/31104207001?pr=211

yeah there was a type error when main was merged in. fixed the issue and made build pass before merging.

mdruker:
Sorry for the unsolicited comment but as an FYI associated flags for some languages can be very politically heated, e.g. a lot of people who use the Russian language who would be offended at an association with the flag of Russia

IsaacMarovitz:
I'm not sure if I like this change from a UX perspective either.

IsaacMarovitz:
> Sorry for the unsolicited comment but as an FYI associated flags for some languages can be very politically heated, e.g. a lot of people who use the Russian language who would be offended at an association with the flag of Russia

This is an excellent point. Looking through the changeset I can see *several* potentially politically charged changes here.

PR #212, opened 2024-10-03
Fix hcap detail

dholms:
Hcap details were changed from a record format (`{[property]: value}`) to an array format (`{property, value}[]`).

### DIFF
--- a/components/repositories/AccountView.tsx
+++ b/components/repositories/AccountView.tsx
@@ -511,7 +511,6 @@ function Details({
   const hcapDetail = Array.isArray(repo.hcaptchaDetails)
     ? (repo.hcaptchaDetails as { property: string; value: string }[])
     : undefined
-  console.log(hcapDetail)
   return (
     <div className="mx-auto mt-6 max-w-5xl px-4 sm:px-6 lg:px-8">
       <dl className="grid grid-cols-1 gap-x-4 gap-y-8 sm:grid-cols-2 mb-10">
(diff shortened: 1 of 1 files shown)


components/repositories/AccountView.tsx
```diff
@@ -534,7 +534,7 @@ function Details({
         )}
         {hcapDetail && (
           <DataField value={ip} label="Hcaptcha">
-            {Object.entries(hcapDetail).map(([property, value]) => (
+            {hcapDetail.map(({ property, value }) => (
```
foysalit, line 537: .map will fail if `hcapDetail` undefined so you can either fallback to `[]` up top or just `?.` here.

components/repositories/AccountView.tsx
```diff
@@ -508,10 +508,10 @@ function Details({
...
     : ''
   const ip = typeof repo.ip === 'string' ? repo.ip : undefined
-  const hcapDetail =
-    typeof repo.hcaptchaDetails === 'object'
-      ? (repo.hcaptchaDetails as Record<string, string>)
-      : undefined
+  const hcapDetail = Array.isArray(repo.hcaptchaDetails)
+    ? (repo.hcaptchaDetails as { property: string; value: string }[])
+    : undefined
+  console.log(hcapDetail)
```
foysalit, line 514: nix the console.log and we're good!

foysalit (changes requested):

components/repositories/AccountView.tsx
dholms: oops 🤦

components/repositories/AccountView.tsx
dholms: ah good looks 👍

foysalit (approved):

PR #213, opened 2024-10-03
Related accounts

dholms:
Building on https://github.com/bluesky-social/atproto/pull/2855

Provide a page on the account view for correlating bad actors such as ban evaders and bot waves across various threat signatures supplied from the entryway. Note that these threat signatures should be obfuscated and privacy-preserving (and are in Bluesky's implementation)


foysalit (approved):
looks good!

PR #220, opened 2024-10-17
:sparkles: Allow adding all lists of a user to workspace

foysalit:
Screenshot 2024-10-17 at 10 12 30


components/list/Lists.tsx
```diff
@@ -21,6 +29,37 @@ export function Lists({ actor }: { actor: string }) {
...
+      do {
+        const nextLists = await labelerAgent.app.bsky.graph.getLists({
+          actor,
+          limit: 25,
+          cursor,
+        })
+        await addItemsToWorkspace(nextLists.data.lists.map((f) => f.uri))
+        cursor = nextLists.data.cursor
+        //   if the modal is closed, that means the user decided not to add any more user to workspace
+      } while (cursor && isConfirmationOpen)
```
matthieusieben, line 55: Due to the nature of how react works, this condition will not actually cancel the loop as `isConfirmationOpen` is a const that will never become false.

For this to work you'll probably need to rely on a ref & AbortController.


```ts
const abortRef = ref<AbortController | null>(null)
const isWorking = abortRef.current != null

const startWorking = async () => {
  // Abort previous run, if any
  abortRef.current?.abort()

  const controller = new AbortController()
  abortRef.current = controller

  try {
    do {
      await someAsyncStuff({
        signal: controller.signal
      })
      
      if (controller.signal.aborted) {
        // make sure the catch block is triggered
        throw controller.signal.reason
      }
    } while (true)
  } catch (err) {
    if (controller.signal.aborted) {
      toast.info('this action was aborted')
    } else {
      toast.error(`Something went wrong: ${(e as Error).message}`)
    }
  } finally {
    // Prevent re-use (for good measure)
    controller.abort()

    // If this run is still the currently active one
    if (abortRef.current === controller) {
      // Cleanup reference
      abortRef.current = null

      // Close the modal
      setIsConfirmationOpen(false)
    }
  } 
}

const abortCurrentRun = () => {
  // No need to call setIsConfirmationOpen here as this will be done through the async process
  abortRef.current?.abort()
}

// Abort the current run when the component is unmounted (e.g. due to navigation)
useEffect(() => {
  return () => abortRef.current?.abort()
}, [abortRef])

```

PS: If this feels too manual, the mutation lib probably already does all of the above

matthieusieben (changes requested):

components/list/Lists.tsx
foysalit: yeah I knew it wasn't optimal but it does kill the requests if other conditions cause a re-render.
we use this pattern in a few places so I cleaned up all usage of this pattern to use abort controllers. thank you for pointing it out!

app/repositories/page-content.tsx
```diff
@@ -112,27 +122,42 @@ function useSearchResultsQuery(q: string) {
       return
     }
     setIsAdding(true)
+    const newAbortController = new AbortController()
```
matthieusieben, line 125: For good measure, we should ensure that only one run is possible at any given time.

```suggestion
    // allready running
    if (abortController.current) return

    const newAbortController = new AbortController()
```
or

```suggestion
    // user cancelled by confirming again
    abortController.current?.abort('user-cancelled')
    const newAbortController = new AbortController()
```

matthieusieben (approved):
These changes all look very similar. I wonder if there is a way to avoid the code duplication by making some kind of hook ? Looks good otherwise.

app/repositories/page-content.tsx
```diff
@@ -90,6 +99,7 @@ const getRepos =
   }
 
 function useSearchResultsQuery(q: string) {
+  const abortController = useRef<AbortController | null>(null)
```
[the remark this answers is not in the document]
foysalit: I think i was avoiding this earlier on so that folks can run one of these and leave it running in the background while they browse and do other things. but in hindsight, that may be a bit too risky since it will make it very easy for people to leave something like this running for 20K followers or something.

app/repositories/page-content.tsx
foysalit: hmm... not sure how this case could happen since every interaction aborts the controller after it is instantiated. especially with the abort on unmount. I will give it another look when refactoring these into a hook.

app/repositories/page-content.tsx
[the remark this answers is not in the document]
foysalit: agreed on the refactoring into a hook recommendation. I'm adding a ticket for that and don't wanna bite that off in this PR. let me get this out and will handle the refactoring in a future, separate PR.

app/repositories/page-content.tsx
matthieusieben: It shouldn't happen, but only because the `isAdding` UI prevents it to. So if someone else breaks that code (eg. by changing the UI), it might end-up happening in parallel. Adding this is just a safety measure against future changes.

PR #232, opened 2024-11-14
:bug: Use profile createdAt timestamp to determine account age

foysalit:


components/subject/RecordAuthorStatus.tsx
```diff
@@ -12,12 +13,21 @@ import {
 } from '@heroicons/react/24/solid'
 import { differenceInDays, formatDistance } from 'date-fns'
 
+const getProfileFromRepo = (repo: ToolsOzoneModerationDefs.RepoView) => {
+  const profile = repo.relatedRecords.find(
+    (r) => r?.['$type'] === CollectionId.Profile,
+  )
+  return profile ? (profile as AppBskyActorProfile.Record) : null
+}
```
matthieusieben, line 21: By using the `isX` type utils, you should be able to avoid having to type cast here


```suggestion
  return repo.relatedRecords.find(AppBskyActorProfile.isRecord)
}
```

matthieusieben (approved):
LGTM

components/subject/RecordAuthorStatus.tsx
foysalit: ooh TIL!

PR #235, opened 2024-11-15
:sparkles: Show warning for high profile accounts based on follower count

foysalit:
This PR adds a prominently visible warning when actioning an account with follower count more than a configured value (through env var)

Screenshot 2024-11-15 at 20 41 00


app/actions/ModActionPanel/QuickAction.tsx
```diff
@@ -828,22 +834,37 @@ function Form(
 function useSubjectQuery(subject: string) {
```
matthieusieben, line 834: nit, and probably a more general comment, but I think that the get profile part would be better implemented as its own `useSubjectProfile` hook. This would allow for better composability and maintainability in the future.

components/repositories/HighProfileWarning.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+  if (!followersCount || followersCount < HIGH_PROFILE_FOLLOWER_THRESHOLD) {
+    return null
+  }
+
+  return (
+    <Alert
+      type="warning"
+      title="High profile account"
+      body={`This user has more than ${HIGH_PROFILE_FOLLOWER_THRESHOLD} followers. Please take caution when moderating this account.`}
```
matthieusieben, line 20: Should the actual count be displayed here (since we have the info at hand) ?

matthieusieben (approved):
Pretty straight forward 👍

app/actions/ModActionPanel/QuickAction.tsx
foysalit: yeah that's fair. just don't have many places here where profile is needed in isolation so feels more natural for me to tie it to repo/record data fetcher for now at least.

PR #241, opened 2024-11-20
Prevent use of another user's server-config

matthieusieben:
The server config is stored in the local storage to allow cached access when loading the app. However, the cache key is the same for every user.

This PR adds a `useCachedQuery` that combines the `useQuery` and `useLocalStorage` hooks using a consistent cache key mechanism for both the cache and query key.

### DIFF
--- a/components/shell/ConfigurationContext/useServerConfigQuery.ts
+++ b/components/shell/ConfigurationContext/useServerConfigQuery.ts
@@ -1,71 +1,74 @@
 import { Agent } from '@atproto/api'
 import { ResponseType, XRPCError } from '@atproto/xrpc'
-import { useQuery } from '@tanstack/react-query'
-import { useEffect } from 'react'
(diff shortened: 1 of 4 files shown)


components/shell/ConfigurationContext/useServerConfigQuery.ts
```diff
@@ -1,71 +1,74 @@
...
-        const reset = Number(error.headers['ratelimit-reset']) * 1e3
-        return reset - Date.now()
-      }
-
-      // Exponential backoff with a maximum of 30 seconds
-      return Math.min(1000 * 2 ** attempt, 30000)
-    },
-    queryKey: ['server-config', agent.assertDid, agent.proxy],
+  return useStoredQuery({
+    queryKey: ['server-config', agent.assertDid, agent.proxy ?? null],
```
matthieusieben, line 9: This is the real fix from this PR

PR #242, opened 2024-11-20
:sparkles: Add protected tags config tab

foysalit:
Depends on https://github.com/bluesky-social/atproto/pull/3050


foysalit:
reviewed this pairing with @matthieusieben earlier and the only change he requested was addressed in the last commit. going to merge but happy to address anything as a followup.

PR #245, opened 2024-11-21
:bug: Build correct event filter for appeal and tags

foysalit:
This PR fixed event filters so that tag based events and appeal event filter work correctly. These are pseudo event types so there's no 1-to-1 mapping for these and the actual filter need to be built manually.

### DIFF
--- a/components/common/labels/Selector.tsx
+++ b/components/common/labels/Selector.tsx
@@ -10,7 +10,7 @@ const EMPTY_ARR = []
 export const LabelSelector = (props: LabelsProps) => {
   const {
     id,
-    formId,
+    form,
     name,
     defaultLabels = EMPTY_ARR,
     disabled,
@@ -57,7 +57,7 @@ export const LabelSelector = (props: LabelsProps) => {
       <input
         type="hidden"
         name={name}
-        {...{ id, formId, disabled }}
+        {...{ id, form, disabled }}
         value={Array.isArray(selectedLabels) ? selectedLabels.join(',') : ''}
       />
       <Input
@@ -114,7 +114,7 @@ export const LabelSelector = (props: LabelsProps) => {
 
 type LabelsProps = {
   id: string
-  formId: string
+  form: string
   name: string
   disabled?: boolean
   defaultLabels?: string[]
--- a/components/mod-event/useModEventList.tsx
+++ b/components/mod-event/useModEventList.tsx
@@ -228,9 +228,7 @@ export const useModEventList = (
           // There is a no appeal type, it's a placeholder and behind the scene
           // we use type as report and reportTypes as appeal
           if (type === MOD_EVENTS.APPEAL) {
-            if (!queryParams.reportTypes) {
-              queryParams.reportTypes = []
-            }
+            queryParams.reportTypes ||= []
             queryParams.reportTypes.push(ComAtprotoModerationDefs.REASONAPPEAL)
             return MOD_EVENTS.REPORT
           }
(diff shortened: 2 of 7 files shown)


components/common/labels/Selector.tsx
```diff
@@ -10,7 +10,7 @@ const EMPTY_ARR = []
 export const LabelSelector = (props: LabelsProps) => {
   const {
     id,
-    formId,
+    form,
```
foysalit, line 13: just cleaning these up.

components/mod-event/useModEventList.tsx
```diff
@@ -321,3 +354,39 @@ export const useModEventList = (
...
+  tagTypes.forEach((key) => {
+    if (type === key) {
+      if (TagBasedTypeFilters[key].add) {
+        add.push(TagBasedTypeFilters[key].add)
+      }
+      if (TagBasedTypeFilters[key].remove) {
+        remove.push(TagBasedTypeFilters[key].remove)
+      }
+    }
+  })
```
matthieusieben, line 389: Looks like that iteration is not really needed:

```suggestion
  if (!(type in tagTypes) || !Object.hasOwn(tagTypes, type)) {
    return { add, remove }
  }
  if (TagBasedTypeFilters[type].add) {
    add.push(TagBasedTypeFilters[type].add)
  }
  if (TagBasedTypeFilters[type].remove) {
    remove.push(TagBasedTypeFilters[type].remove)
  }
```

components/mod-event/useModEventList.tsx
```diff
@@ -321,3 +354,39 @@ export const useModEventList = (
...
+  [MOD_EVENTS.ENABLE_DMS]: { remove: DM_DISABLE_TAG, add: '' },
+  [MOD_EVENTS.DISABLE_VIDEO_UPLOAD]: {
+    add: VIDEO_UPLOAD_DISABLE_TAG,
+    remove: '',
+  },
+  [MOD_EVENTS.ENABLE_VIDEO_UPLOAD]: {
+    remove: VIDEO_UPLOAD_DISABLE_TAG,
+    add: '',
+  },
+}
```
matthieusieben, line 369: Using objects as a map is not ideal in javascript (because `TagBasedTypeFilters['toString']` will be there and might cause abuse). You can make it safer by using a `Map` (best) or by removing the inheritance of `Object.prototype` like so (technically safe but funkier in terms of maintainability):

```suggestion
const TagBasedTypeFilters = {
  __proto__: null, // Allow using this as a "map"
  [MOD_EVENTS.DISABLE_DMS]: { add: DM_DISABLE_TAG, remove: '' },
  [MOD_EVENTS.ENABLE_DMS]: { remove: DM_DISABLE_TAG, add: '' },
  [MOD_EVENTS.DISABLE_VIDEO_UPLOAD]: {
    add: VIDEO_UPLOAD_DISABLE_TAG,
    remove: '',
  },
  [MOD_EVENTS.ENABLE_VIDEO_UPLOAD]: {
    remove: VIDEO_UPLOAD_DISABLE_TAG,
    add: '',
  },
}
```

components/mod-event/useModEventList.tsx
```diff
@@ -225,6 +222,42 @@ export const useModEventList = (
...
 
+      const filterTypes = types.filter(Boolean)
+      if (filterTypes.length < allTypes.length && filterTypes.length > 0) {
+        queryParams.types = filterTypes.map((type) => {
+          // There is a no appeal type, it's a placeholder and behind the scene
+          // we use type as report and reportTypes as appeal
+          if (type === MOD_EVENTS.APPEAL) {
+            if (!queryParams.reportTypes) {
+              queryParams.reportTypes = []
+            }
```
matthieusieben, line 233: Nit: I personnaly like using the following notation.

```suggestion
            queryParams.reportTypes ||= []
```

components/mod-event/useModEventList.tsx
```diff
@@ -225,6 +222,42 @@ export const useModEventList = (
...
+
+          // We use custom event type name that translate to either add or remove certain tag
+          const { add, remove } = buildTagFilter(type)
+          if (add.length || remove.length) {
+            if (add.length) {
+              if (queryParams.addedTags) {
+                queryParams.addedTags.push(...add)
+              } else {
+                queryParams.addedTags = add
+              }
```
matthieusieben, line 246: Nit: the `||=` notation make things easier to read imo

```suggestion
              queryParams.addedTags ||= []
              queryParams.addedTags.push(...add)
```

matthieusieben (approved):
Looks good. Only stylistic comments. 👍

matthieusieben:
There seems to be an issue here https://github.com/bluesky-social/ozone/pull/245/files/c097b03a05ffabc6531e0dc252d9ca8ae5d60fe0#file-app-actions-modactionpanel-quickaction-tsx-L667

components/mod-event/useModEventList.tsx
foysalit: nice thanks!

components/mod-event/useModEventList.tsx
foysalit: hmm... agreed maps are better but not convinced we should be worried about this type of "abuse" in this context.

PR #247, opened 2024-11-23
:sparkles: Add people tab on search page

foysalit:
This PR adds a new `People` tab on the search page similar to the app's search page which shows all users matching the search query.

### DIFF
--- a/components/search-content/useContentSearch.tsx
+++ b/components/search-content/useContentSearch.tsx
@@ -1,22 +1,53 @@
-import { useLabelerAgent } from '@/shell/ConfigurationContext'
+import { useAppviewAgent, useLabelerAgent } from '@/shell/ConfigurationContext'
 import { useWorkspaceAddItemsMutation } from '@/workspace/hooks'
+import {
+  AppBskyActorDefs,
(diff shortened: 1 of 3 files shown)


components/search-content/useContentSearch.tsx
```diff
@@ -1,22 +1,53 @@
...
   const { mutate: addToWorkspace } = useWorkspaceAddItemsMutation()
 
-  const searchInfo = useInfiniteQuery({
+  const searchInfo = useInfiniteQuery<QueryData<typeof section>>({
     queryKey: ['searchContent', { term, section }],
     enabled: !!term,
     queryFn: async ({ pageParam }) => {
+      if (section === 'people') {
+        // This is actually never hit, we always have appview configured
+        if (!appviewAgent) {
```
matthieusieben, line 36: Can this really be null?!

matthieusieben (approved):

components/search-content/useContentSearch.tsx
foysalit: nope. mostly for bypassing the ts check. technically, without appview url configured, it can be null

PR #249, opened 2024-11-29
:sparkles: Load and display moderation status in event list

foysalit:
This PR hydrates repo and record details on the client side for each event and shows the moderation status of the subject of when viewing event stream.

Screenshot 2024-11-29 at 23 08 52


components/mod-event/EventItem.tsx
```diff
@@ -102,12 +103,19 @@ function isMessageSubject(
   return subject.messageId !== undefined
 }
 
+type ModEventType<T> = { event: T } & ToolsOzoneModerationDefs.ModEventView
```
foysalit, line 106: Unrelated cleanup to get rid of the ts-ignores

PR #252, opened 2024-12-03
:lipstick: Hide blob list by default and add image viewer

foysalit:
To make more room in the Quick Action Panel, we are now hiding blob list by default, allowing mods to open it as a dropdown when they actually want to action blobs.

Also added an image viewer to the blob list so that mods can view each individual blob item.


app/actions/ModActionPanel/BlobList.tsx
```diff
@@ -1,25 +1,36 @@
...
+      {!blobs.length ? (
+        <div className="text-sm text-gray-400">No blobs found</div>
+      ) : (
+        <BlobListLightbox
+          blobs={blobs}
+          authorDid={props.authorDid}
+          slideIndex={lightboxImageIndex}
+          onClose={() => setLightboxImageIndex(-1)}
+        />
+      )}
```
matthieusieben, line 31: Nit: for the sake of readability I would invert the conditions `blobs.length > 0 ? <BlobListLightbox /> : <div />`

components/common/BlobListLightbox.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { useCallback } from 'react'
+import Lightbox from 'yet-another-react-lightbox'
+import Captions from 'yet-another-react-lightbox/plugins/captions'
+import Thumbnails from 'yet-another-react-lightbox/plugins/thumbnails'
+
+export const useBlobUrl = ({ authorDid }: { authorDid: string }) => {
+  const { appview } = useServerConfig()
+  const cdnUrl = appview?.endsWith('.bsky.app')
+    ? 'https://cdn.bsky.app'
+    : appview
```
matthieusieben, line 12: Wouldn't it be better to make that part of the server config ?

components/common/BlobListLightbox.tsx
```diff
@@ -0,0 +1,89 @@
...
+    }
+  }
+
+  console.log(
+    blobs.map((blob) => ({
+      src: getBlobUrl({
+        cid: blob.cid,
+      }),
+    })),
+  )
```
matthieusieben, line 61:
```suggestion
```

components/common/BlobListLightbox.tsx
```diff
@@ -0,0 +1,89 @@
...
+  blobs: ToolsOzoneModerationDefs.BlobView[]
+  onClose: () => void
+  slideIndex: number
+}) => {
+  const getBlobUrl = useBlobUrl({ authorDid })
+  const handleKeyDown = (event) => {
+    if (event.key === 'Escape') {
+      event.stopPropagation()
+    }
+  }
```
matthieusieben, line 53: You need to `useCallback` here. Otherwise the `document.addEventListener('keydown', handleKeyDown)` and `document.removeEventListener('keydown', handleKeyDown)` might not refer to the same function (resulting in the listener staying there forever)

matthieusieben (approved):
Looks good overall. Some small comments

components/common/BlobListLightbox.tsx
foysalit: yep but im trying to reduce server side touch points. i have a ticket on trello to send it from server.

components/common/BlobListLightbox.tsx
foysalit: ermm.. not sure if that's correct? what would the deps for the hook be?

foysalit: happy to address this later and if we find this to be an issue, this will need to be updated in a few other places where lightbox is used.

PR #254, opened 2024-12-03
:sparkles: Optionally ack all subjects of an account when acknowledgi…

foysalit:
Depends on https://github.com/bluesky-social/atproto/pull/3171

https://github.com/user-attachments/assets/c4de8d82-3baa-4e33-81e4-5cee27071409

### DIFF
--- a/app/repositories/page-content.tsx
+++ b/app/repositories/page-content.tsx
@@ -1,12 +1,13 @@
 import { SectionHeader } from '../../components/SectionHeader'
 import { RepositoriesTable } from '@/repositories/RepositoriesTable'
-import { useSearchParams } from 'next/navigation'
+import { usePathname, useRouter, useSearchParams } from 'next/navigation'
 import { useInfiniteQuery } from '@tanstack/react-query'
 import { useTitle } from 'react-use'
 import {
   Agent,
   ToolsOzoneModerationDefs,
   ComAtprotoAdminSearchAccounts,
+  ToolsOzoneModerationEmitEvent,
 } from '@atproto/api'
 import { useLabelerAgent } from '@/shell/ConfigurationContext'
 import { ActionButton } from '@/common/buttons'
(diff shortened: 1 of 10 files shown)


components/mod-event/ItemTitle.tsx
```diff
@@ -119,11 +124,8 @@ export const ItemTitle = ({
             hideActor={!showContentAuthor}
             subjectRepoHandle={modEvent.subjectHandle}
           />
-          {modEvent.repo?.moderation.subjectStatus && (
-            <ReviewStateIcon
-              size="sm"
-              subjectStatus={modEvent.repo.moderation.subjectStatus}
-            />
+          {subjectStatus && (
+            <ReviewStateIcon size="sm" subjectStatus={subjectStatus} />
```
foysalit, line 128: Bugfix unrelated to the PR

app/repositories/page-content.tsx
```diff
@@ -1,12 +1,13 @@
 import { SectionHeader } from '../../components/SectionHeader'
 import { RepositoriesTable } from '@/repositories/RepositoriesTable'
-import { useSearchParams } from 'next/navigation'
```
foysalit, line 3: Changes in this file are unrelated to the PR but I was working around this area so added these on. 

1. show loading state on repo list table
2. use repo moderation states to show subject overview and subject status icon just like we do in the queue page instead of having it show just the handle in the first column and subject status in the last column.
3. add quick action panel in repositories page so that mods can directly action accounts from the page without having to go to profile screen first.
3. improve moderation status loading performance by bulk fetching using getRepos instead of calling getRepo on every did.

matthieusieben (approved):
nice !

SteveTownsend:
very cool, @nafo-moderation.bsky.social will get a lot of use out of this since we aggressively label accounts, and now can auto-ack the child reports

PR #256, opened 2024-12-05
:sparkles: Allow managing page size for event stream

foysalit:
When investigating and actioning large blocks of accounts/posts, mods often need to look at event stream in depth and being limited to 25 items per page is quite annoying. However, for quick action panel and other places, 25 items are more than enough for most cases. So making the page size configurable gets us the best of both worlds.

https://github.com/user-attachments/assets/6808e3c4-fe71-43f0-b3c7-2934a73a3e39


matthieusieben (approved):
nice

PR #259, opened 2024-12-11
✨ Bulk email from workspace

foysalit:
This PR adds email event to both quick action panel and workspace panel which allows moderators to send emails directly from quick action panel after a takedown or similar action and in bulk from workspace.

https://github.com/user-attachments/assets/812d9a6e-a4f0-47c8-b7c3-6f00f9526900


app/actions/ModActionPanel/QuickAction.tsx
```diff
@@ -507,299 +535,309 @@ function Form(
...
-              <FormLabel
-                label="Subject"
-                htmlFor="subject"
-                className="flex-1"
-                copyButton={{ text: subject, label: 'Copy subject' }}
-                extraLabel={
-                  <SubjectSwitchButton
-                    subject={subject}
-                    setSubject={setSubject}
+        <div className="flex sm:w-1/2 flex-col">
```
foysalit, line 538: The diff below looks scary because of this root level wrapper element change. Almost nothing changed below except for a level of indentation and positioning in line.

package.json
```diff
@@ -52,7 +52,7 @@
     "@tailwindcss/forms": "^0.5.3",
     "@tailwindcss/typography": "^0.5.8",
     "autoprefixer": "^10.4.13",
-    "cypress": "^12.14.0",
+    "cypress": "^13.16.1",
```
foysalit, line 55: this is not related to the PR of course but i don't touch these tests often so wanted to take this opportunity to upgrade

components/email/Composer.tsx
```diff
@@ -93,31 +105,37 @@ export const EmailComposer = ({ did }: { did: string }) => {
...
-              return 'Error sending email'
+            error: {
+              render() {
+                return 'Error sending email'
+              },
             },
           },
-        },
-      )
+        )
```
matthieusieben, line 136: Could we make `handleSubmit` non-optional and use this implementation as `handleSubmit` wherever this is used ? Alternatively, a simpler re-factor would be to create a new component that provides this `handleSubmit` implementation:

```tsx
export function EmailComposer ({ did }: { did: string }) {
  return <EmailComposerBase
    did={did}
    handleSubmit={
      // tools.ozone.moderation.emitEvent + toast here
    }
    />
})

export function EmailComposerBase ({
  did,
  replacePlaceholders = true,
  handleSubmit,
}: {
  did?: string
  replacePlaceholders?: boolean
  handleSubmit?: (emailData: EmailComposerData) => Promise<void>
}) {

}
```

components/mod-event/helpers/emitEvent.tsx
```diff
@@ -101,13 +148,14 @@ const emitEventsInBulk = async ({
...
-          await labelerAgent.api.tools.ozone.moderation.emitEvent({
+          await labelerAgent.tools.ozone.moderation.emitEvent({
+            ...eventForSubject(eventData, subjectData[sub]),
             subject,
             createdBy: labelerAgent.assertDid,
-            ...eventData,
           })
           results.succeeded.push(sub)
         } catch (err) {
+          console.error(err)
```
matthieusieben, line 158:
```suggestion
```

components/workspace/Panel.tsx
```diff
@@ -50,13 +51,16 @@ export function WorkspacePanel(props: PropsOf<typeof ActionPanel>) {
...
   const dropzoneRef = createRef<DropzoneRef>()
 
-  const handleRemoveSelected = () => {
-    const selectedItems = Array.from(
+  const getSelectedItems = () => {
+    return Array.from(
       formRef.current?.querySelectorAll<HTMLInputElement>(
         'input[type="checkbox"][name="workspaceItem"]:checked',
       ) || [],
     ).map((checkbox) => checkbox.value)
```
matthieusieben, line 59: Did you know: the second argument of `Array.from` is a map function ;-)

```suggestion
    return Array.from(
      formRef.current?.querySelectorAll<HTMLInputElement>(
        'input[type="checkbox"][name="workspaceItem"]:checked',
      ) || [],
      (checkbox) => checkbox.value
    )
```

app/actions/ModActionPanel/QuickAction.tsx
matthieusieben: Maybe not would be a good time to actually re-factor this huge component into smaller ones ?

matthieusieben (approved):
Look fine 👍

app/actions/ModActionPanel/QuickAction.tsx
foysalit: yeah it just keeps growing but honestly, this is like the most used and impactful piece of code in the whole app so kinda nervous to do that along with a PR. let me do a separate PR for it.

components/workspace/Panel.tsx
foysalit: nope. TIL!

foysalit:
refactored PR coming up soon.

PR #260, opened 2024-12-12
Add account details to queue

matthieusieben:


app/reports/page-content.tsx
```diff
@@ -311,6 +311,8 @@ function useModerationQueueQuery() {
     useFluentReportSearchParams()
 
   return useInfiniteQuery({
+    // lighten load on the server
```
foysalit, line 314: I think this made sense when we had almost no volume compared to now. however so im up for changing this but I worry if mods are too dependent on auto reload now and will forget to refresh when they actually want to get latest items.

foysalit (approved):

PR #263, opened 2024-12-21
:sparkles: Pass queue config to the server query when fetching statuses

foysalit:
Depends on https://github.com/bluesky-social/atproto/pull/3280

This PR removes the client side queue splitting and passes the queue config to the request when calling `queryStatuses` so that the db query on the server can take care of the splitting.

### DIFF
--- a/components/config/Labeler.tsx
+++ b/components/config/Labeler.tsx
@@ -41,8 +41,8 @@ export function LabelerConfig() {
       )}
 
       <ServerConfig />
-      <LocalPreferences />
       <QueueSetting />
+      <LocalPreferences />
       <ExternalLabelerConfig />
     </div>
   )
--- a/components/config/LocalPreferences.tsx
+++ b/components/config/LocalPreferences.tsx
@@ -66,7 +66,7 @@ export const LocalPreferences = () => {
           You can choose to make media content (video and image) with the
           following labels appear on your screen with your preferred filter.
           <br />
-          This is your personal configuration and won't be shared with other
+          This is your personal configuration and won{"'"}t be shared with other
           moderators.
         </p>
 
--- a/components/config/LocalPreferences.tsx
+++ b/components/config/LocalPreferences.tsx
@@ -65,6 +65,9 @@ export const LocalPreferences = () => {
         <p className="text-sm mb-2">
           You can choose to make media content (video and image) with the
           following labels appear on your screen with your preferred filter.
+          <br />
+          This is your personal configuration and won't be shared with other
+          moderators.
         </p>
 
         <form onSubmit={handleSubmit}>
--- a/components/config/external-labeler.tsx
+++ b/components/config/external-labeler.tsx
@@ -46,7 +46,9 @@ export const ExternalLabelerConfig = () => {
             subscriptions there.
           </p>
           <p className="mt-1 text-sm text-gray-900 dark:text-gray-200">
-            You can unsubscribe from any external labeler at any time.
+            You can unsubscribe from any external labeler at any time. This is
+            your personal configuration and won't be shared with other
+            moderators.
           </p>
 
           <div className="flex flex-row justify-end items-end my-3 gap-2">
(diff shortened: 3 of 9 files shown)


components/config/LocalPreferences.tsx
```diff
@@ -65,6 +65,9 @@ export const LocalPreferences = () => {
         <p className="text-sm mb-2">
           You can choose to make media content (video and image) with the
           following labels appear on your screen with your preferred filter.
+          <br />
```
foysalit, line 68: Not related to the PR but adding additional clarification as requested by mods.

components/config/external-labeler.tsx
```diff
@@ -46,7 +46,9 @@ export const ExternalLabelerConfig = () => {
             subscriptions there.
           </p>
           <p className="mt-1 text-sm text-gray-900 dark:text-gray-200">
-            You can unsubscribe from any external labeler at any time.
+            You can unsubscribe from any external labeler at any time. This is
```
foysalit, line 49: Not related to the PR but adding additional clarification as requested by mods.

components/config/Labeler.tsx
```diff
@@ -41,8 +41,8 @@ export function LabelerConfig() {
       )}
 
       <ServerConfig />
-      <LocalPreferences />
```
foysalit, line 44: Not related to the PR but relocating the panels as requested by mods.

components/mod-event/EventItem.tsx
```diff
@@ -31,6 +31,38 @@ const LinkToAuthor = ({
...
+          {part}
+        </a>
+      )
+    }
+    // Otherwise, return it as plain text
+    return <span key={index}>{part}</span>
+  })
+}
+
+const TextWithLinks: React.FC<{ text: string }> = ({ text }) => {
```
foysalit, line 62: Not related to the PR but mods reported that sometimes they leave links to verification/investigation content and the links always cause the layout to break. This is basically looking for links in comment and making sure they are used wrapped in a tag and wraps the text properly to next line.

components/mod-event/EventItem.tsx
```diff
@@ -31,6 +31,38 @@ const LinkToAuthor = ({
...
+// Utility function to detect and replace links with <a> tags
+const wrapLinksInText = (text: string): JSX.Element[] => {
+  // Regular expression to match URLs
+  const urlRegex = /(https?:\/\/[^\s]+)/g
+
+  // Split text into parts, with URLs as matches
+  const parts = text.split(urlRegex)
+
+  return parts.map((part, index) => {
+    if (urlRegex.test(part)) {
```
devinivy, line 43: Global (`/g`) regexes are stateful when used with `.test()`, taking into account the last matched index.  I think that this could lead to missed matches here.

components/mod-event/EventItem.tsx
```diff
@@ -31,6 +31,38 @@ const LinkToAuthor = ({
...
 
+// Utility function to detect and replace links with <a> tags
+const wrapLinksInText = (text: string): JSX.Element[] => {
+  // Regular expression to match URLs
+  const urlRegex = /(https?:\/\/[^\s]+)/g
+
+  // Split text into parts, with URLs as matches
+  const parts = text.split(urlRegex)
+
+  return parts.map((part, index) => {
```
devinivy, line 42: The parts matching the regex wont be included in `parts`— I think this will effectively filter out links.

components/mod-event/EventItem.tsx
devinivy: For example:
```js
const regex = /a/g
regex.test('a') // true
regex.test('a') // false
```

devinivy (approved):

components/mod-event/EventItem.tsx
foysalit: my regex-foo is not good at all but stackoverflow said wrapping it in parentheses makes it a capture group which is why the matches will be included in the returned array.

components/mod-event/EventItem.tsx
foysalit: again, not sure how accurate I am with this but it seems to be only an issue when applied to same string and since we're applying it on different `parts` I think this works fine...? I did test this with various possible inputs where it may fail but it seems to work fine but happy to make changes if we find issues down the road.
Screenshot 2024-12-23 at 22 43 27

PR #276, opened 2025-01-09
:sparkles: Add self-harm to list of configurable media filters

foysalit:


foysalit:
one liner requested by mods, not bugging folks for a review on this.

PR #279, opened 2025-01-16
:sparkles: Show subject stats/summary

foysalit:
Depends on https://github.com/bluesky-social/atproto/pull/3236

This PR shows moderation stats/summary for subjects/accounts.

Screenshot 2025-01-21 at 16 18 14

> Stats preview in the queue table

Screenshot 2025-01-21 at 16 18 24

> Sort config for stats

Screenshot 2025-01-21 at 16 18 45

> Full stats view in quick action panel

Screenshot 2025-01-21 at 16 18 52

> Stats summary in quick action panel

Screenshot 2025-01-21 at 17 44 52

> Stats filter view

### DIFF
--- a/components/reports/useQueueFilter.tsx
+++ b/components/reports/useQueueFilter.tsx
@@ -123,6 +123,18 @@ export const useQueueFilter = () => {
     updateFilters({ tags: [] })
   }
 
+  const setMinAccountSuspendCount = (minAccountSuspendCount?: number) => {
+    updateFilters({ minAccountSuspendCount })
+  }
+
+  const setMinReportedRecordsCount = (minReportedRecordsCount?: number) => {
+    updateFilters({ minReportedRecordsCount })
+  }
+
+  const setMinTakendownRecordsCount = (minTakendownRecordsCount?: number) => {
+    updateFilters({ minTakendownRecordsCount })
+  }
+
   return {
     queueFilters,
     updateFilters,
@@ -132,5 +144,8 @@ export const useQueueFilter = () => {
     toggleSubjectType,
     clearSubjectType,
     clearTags,
+    setMinAccountSuspendCount,
+    setMinReportedRecordsCount,
+    setMinTakendownRecordsCount,
   }
 }
--- a/components/subject/Summary.tsx
+++ b/components/subject/Summary.tsx
@@ -142,8 +142,10 @@ export const AccountStats = ({
 export const RecordsStats = ({
   stats,
   showAll,
+  onlyNumbers,
 }: {
   showAll: boolean
+  onlyNumbers?: boolean
   stats: ToolsOzoneModerationDefs.RecordsStats
 }) => {
   const {
@@ -153,7 +155,7 @@ export const RecordsStats = ({
     appealedCount,
     pendingCount,
     processedCount,
-    takedownCount,
+    takendownCount,
   } = stats
 
   const hasStats =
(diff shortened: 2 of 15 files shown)


components/subject/table.tsx
```diff
@@ -81,6 +91,55 @@ export function SubjectTable(
   )
 }
 
+const SubjectSummaryColumn = ({
```
foysalit, line 94: for now, we don't want to expose all stats we have. chatted with mod team and agreed that these stats may impact decisions in unwanted ways.

components/reports/useQueueFilter.tsx
```diff
@@ -123,6 +123,18 @@ export const useQueueFilter = () => {
...
+    updateFilters({ minAccountSuspendCount })
+  }
+
+  const setMinReportedRecordsCount = (minReportedRecordsCount?: number) => {
+    updateFilters({ minReportedRecordsCount })
+  }
+
+  const setMinTakendownRecordsCount = (minTakendownRecordsCount?: number) => {
+    updateFilters({ minTakendownRecordsCount })
+  }
```
matthieusieben, line 136: Nit: It feels like exposing both `updateFilters` and a helper that wraps `updateFilters` can be confusing when using the API. Couldn't we just require users of this hook to do:

```ts
const { updateFilters } = useQuereFilter()

updateFilters({ minTakendownRecordsCount: value })
```

components/subject/Summary.tsx
```diff
@@ -0,0 +1,283 @@
...
+  stats: ToolsOzoneModerationDefs.AccountStats
+}) => {
+  const { takedownCount, suspendCount, appealCount, reportCount } = stats
+  // Dont want to make the UI noisy with appeal count if we're in summary mode and already showing takedown and suspension
+  const shouldShowAppeal =
+    appealCount && (showAll || !takedownCount || !suspendCount)
+  // Dont want to make the UI noisy with report count if we're in summary mode and we have both takedown and suspension count
+  const shouldShowReport =
+    reportCount && (showAll || (!takedownCount && !suspendCount))
+  const hasStats = takedownCount || suspendCount || appealCount || reportCount
```
matthieusieben, line 72: IMO A `zero` should count as "has stats". But if that's really not what you want, this could should explicitly account for `0` vs `undefined`.

```suggestion
  const { takedownCount, suspendCount, appealCount, reportCount } = stats

    // Dont want to make the UI noisy with appeal count if we're in summary mode and already showing takedown and suspension
  const shouldShowAppeal = appealCount == null
    ? true // No data <== I suggest we *do* show that we don't have data
    : showAll ||
      !takedownCount || // No data or zero
      !suspendCount // No data or zero

  // Dont want to make the UI noisy with report count if we're in summary mode and we have both takedown and suspension count
  const shouldShowReport = reportCount == null
    ? true // no data <== 
    : showAll || (!takedownCount && !suspendCount)

  const hasStats =
    takedownCount != null ||
    suspendCount != null ||
    appealCount != null ||
    reportCount != null
```

components/subject/Summary.tsx
```diff
@@ -0,0 +1,283 @@
...
+              : ''
+          }
+          title={`This account was reported ${pluralize(reportCount, 'time')}`}
+        />
+      )}
+    </>
+  )
+}
+
+export const RecordsStats = ({
```
matthieusieben, line 142: Again, I suggest we
1) make a clear distinction between `0` and `undefined`, to make the business logic clearer in the code
2) that we do not use `undefined` as a reason not to show something (using a question mark in that case)

matthieusieben (approved):

components/reports/useQueueFilter.tsx
foysalit: agreed but we are not really using the `updateFilters` anywhere outside of the hook so I'll just remove it from the returned object.

components/subject/Summary.tsx
foysalit: generally agreed but I don't think it impacts mod decisions to know that "we don't have stats" vs. the stats is 0 and I'm trying to be cautious about making the UI too noisy with stats.
at least to start with, I see these as additional metadata to help enforce decisions made from other factors already rather than a primary factor for making decision.

PR #283, opened 2025-01-28
:sparkles: Show subject type and collection filter in events filter panel

foysalit:
This PR brings the already available filters for subject type and record collection to the event list allowing moderators to monitor events on account or record of certain collections either from event page or the quick action panel.

### DIFF
--- a/components/reports/QueueFilter/SubjectType.tsx
+++ b/components/reports/QueueFilter/SubjectType.tsx
@@ -21,6 +21,40 @@ export const QueueFilterSubjectType = () => {
     }) || []
 
   return (
+    <SubjectTypeFilter
+      hasSubjectTypeFilter={hasSubjectTypeFilter}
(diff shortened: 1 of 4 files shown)


components/reports/QueueFilter/SubjectType.tsx
```diff
@@ -20,6 +20,40 @@ export const QueueFilterSubjectType = () => {
...
+        !!selectedIncludeEmbedTypes.length
+      }
+      toggleCollection={toggleCollection}
+      toggleSubjectType={toggleSubjectType}
+      clearSubjectType={clearSubjectType}
+    />
+  )
+}
+
+export const SubjectTypeFilter = ({
```
foysalit, line 40: Refactored the UI part of it into its own component so that we can reuse this in queue filter and event list filter.

PR #284, opened 2025-01-30
:sparkles: Paginate events when adding to workspace

foysalit:


components/mod-event/EventList.tsx
```diff
@@ -266,11 +267,17 @@ export const ModEventList = (
 
           <ConfirmationModal
             onConfirm={() => {
-              addToWorkspace().then(() => setShowWorkspaceConfirmation(null))
+              if (!isAddingToWorkspace) {
+                addToWorkspace().then(() => setShowWorkspaceConfirmation(null))
```
matthieusieben, line 271: Don't we need a `finally` instead of a `then` here?

components/mod-event/useModEventList.tsx
```diff
@@ -404,32 +462,39 @@ export const useModEventList = (
...
+            pageParam: cursor,
+          },
+          { signal: abortController.current?.signal },
+        )
+        await addFromEventsToWorkspace(nextPage.events)
+        cursor = nextPage.cursor
+        //   if the modal is closed, that means the user decided not to add any more user to workspace
+      } while (cursor && showWorkspaceConfirmation)
+    } catch (e) {
+      if (abortController.current?.signal.reason === 'user-cancelled') {
```
matthieusieben, line 490: You can also use the `aborted` property, which is the more common way of checking if a signal was aborted.

```suggestion
      if (abortController.current?.signal.aborted) {
```

components/mod-event/useModEventList.tsx
```diff
@@ -241,143 +392,50 @@ export const useModEventList = (
...
-      if (removedLabels.length) {
-        queryParams.removedLabels = removedLabels
-      }
-
-      if (oldestFirst) {
-        queryParams.sortDirection = 'asc'
-      }
+  useEffect(() => {
+    if (!showWorkspaceConfirmation) {
+      abortController.current?.abort('user-cancelled')
```
matthieusieben, line 397: Only `Error` instances should ever be used as abort reason. This is otherwise similar to `throw "some string"`. Note that you can do `ac.abort()` (no argument).

matthieusieben (approved):

PR #294, opened 2025-02-20
:sparkles: Upgrade atproto/api sdk to 0.14.2 and refactor

foysalit:


components/mod-event/EventItem.tsx
```diff
@@ -101,10 +108,12 @@ const Comment = ({
       {/* This is only for legacy actions, new actions won't have these properties for these events */}
       <EventLabels
         header="Added: "
+        // @ts-ignore
```
foysalit, line 111: These are only for legacy support for data that is more than a yr old so not too worried about typings here.

components/common/feeds/PostThread.tsx
```diff
@@ -66,7 +66,7 @@ export function ThreadPost({
         <ThreadPostWrapper depth={depth} highlight={highlight}>
           <PostAsCard
             className="bg-transparent px-3 py-2"
-            item={thread}
+            item={{ post: thread.post }}
```
matthieusieben, line 69: nit: It would probably be "cleaner" to change the props of `PostAsCard` to be:

```
  post: PostView
  reply?: ReplyRef
  reason?: $Typed<ReasonRepost> | $Typed<ReasonPin> | { $type: string }
  feedContext?: string
```

instead of `item: AppBskyFeedDefs.FeedViewPost` (I mean, this here feels kinda odd)

matthieusieben (approved):
Looks good

PR #308, opened 2025-03-03
❇️ Team member search

foysalit:
Depends on https://github.com/bluesky-social/atproto/pull/3587

This PR adds search feature on team member page and uses the newly available search api in other places where full team member list was being loaded to allow searching by handle/name.

Adds some refactors in usage of getProfiles and fixes formatting issue in user info copy feature.

https://github.com/user-attachments/assets/5375b859-1e6c-4987-8fbe-ba04e35a55a2

### DIFF
--- a/components/team/helpers.ts
+++ b/components/team/helpers.ts
@@ -37,3 +37,13 @@ export const isRoleSuperiorOrSame = (manager: string, target: string) => {
 
   return false
 }
+
+export const createTeamPageLink = (queryParams: Record<string, string>) => {
+  const url = new URL(window.location.href.replace(window.location.search, ''))
+
+  Object.entries({ tab: 'members', ...queryParams }).forEach(([key, value]) => {
+    url.searchParams.set(key, value)
+  })
+
+  return url.toString()
+}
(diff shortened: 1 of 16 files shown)


components/team/helpers.ts
```diff
@@ -37,3 +37,13 @@ export const isRoleSuperiorOrSame = (manager: string, target: string) => {
...
 }
+
+export const createTeamPageLink = (queryParams: Record<string, string>) => {
+  const url = new URL(window.location.href.replace(window.location.search, ''))
+
+  Object.entries({ tab: 'members', ...queryParams }).forEach(([key, value]) => {
+    url.searchParams.set(key, value)
+  })
+
+  return url.toString()
```
matthieusieben, line 48: Did you known ?

```suggestion
  const url = new URL(window.location.href)
  
  url.search = new URLSearchParams(Object.entries({ tab: 'members', ...queryParams })).toString()

  return url.toString()
```

foysalit: yeah but this loses any already existing params that may already exist in the url, right? we want to keep all params and only add the ones that queryParams adds.

PR #311, opened 2025-03-04
:sparkles: Add account age filter to workspace

foysalit:
Allow filtering workspace entries by minimum account age

Screenshot 2025-03-04 at 18 53 28


components/workspace/FilterSelector.tsx
```diff
@@ -17,6 +20,22 @@ const toggleItemCheck = (item: string, select: boolean = true) => {
   }
 }
 
+const matchMinimumAccountAge = (
+  minAccountAgeUnit: string,
```
devinivy, line 24: Would it be possible/reasonable to type this as `'days' | 'hours'`?

devinivy (approved):

components/workspace/FilterSelector.tsx
foysalit: we could but it's a bit dependent on typing the whole state object which im not doing right now and im not seeing much benefit of doing that right now.

Issue #310, opened 2025-03-03
Accessibility: colors chosen for background/text report count flags are WCAG violation


rahaeli:
Hooray for the addition of report counts! Unfortunately, the colors chosen are a WCAG contrast violation and also an accessibility problem beyond pure WCAG standards.

Problem: .bg-blue-400 background (RGB 96, 165, 250) with .text-blue-800 text (60, 64, 175) is a contrast ratio of 3.43:1, which fails both WCAG 2.0 standard 1.4.3 level AA (standard size text must be 4.5:1) and WCAG 2.0 standard 1.4.3 level AAA (standard size text must be 7:1). 

The pure "meet this standard only" fix: setting .text-blue-800 to pure black would give you 8.25:1. 

However, using a colored background at all there on the queue view is an accessibility problem that is not part of the WCAG standards yet because the science on it is extremely in its infancy: it turns scrolling the queue view into a photosensitive vertigo migraine trigger. I will skip the full explanation unless you want it because it's long and complicated, but the extremely short version: if blocks of high-contrast color appear and disappear or move as you scroll or page down, it has a high chance of triggering photosensitive migraine in those who are susceptible (yes, this is why "dark mode" works better for people who get photosensitive migraine, because the contrast is usually lower). 

I strongly recommend setting both the foreground and background color on that element in *at least* the queue view to the same foreground and background color as the standard text for every other element in the table row. It's less urgent to change on the report view, because that doesn't flicker in and out when you scroll, but it's probably easier not to specialcase.

rahaeli:
Copying over my comment from the PR: with the new colors, the orange/rust is 5.26:1 (fails WCAG AAA but passes WCAG AA); the light blue/navy is 6.04:1 (also fails WCAG AAA but passes WCAG AA). I don't know how which you're aiming for (AAA is difficult but not impossible) or whether you want to address the photosensitive migraine trigger, so I'll leave this open for confirmation on that!

PR #315, opened 2025-03-07
improve info tag contrast

bnewbold:
Not great, but at least a bit better. My web accessibility dev tool thing isn't detecting/confirming if this is enough.

Top is changed colors, bottom is old colors.

2025-03-06T23:40:53,269131906-08:00

This is related to @rahaeli's https://github.com/bluesky-social/ozone/issues/310

Maybe even better/clearer would be something like `border-2 border-blue-400`? I'm not sure if that is clear enough for any current workflows, and hesitant to make a bigger change w/o mod team feedback.

### DIFF
--- a/components/subject/PriorityScore.tsx
+++ b/components/subject/PriorityScore.tsx
@@ -15,7 +15,7 @@ export const PriorityScore = ({
   return (
     <LabelChip
       className={classNames(
-        'flex flex-row gap-1 items-center bg-orange-300 text-orange-800',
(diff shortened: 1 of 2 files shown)


foysalit (approved):

rahaeli:
The orange/rust is 5.26:1 (fails WCAG AAA but passes WCAG AA); the light blue/navy is 6.04:1 (also fails WCAG AAA but passes WCAG AA). The [WebAIM contrast checker](https://webaim.org/resources/contrastchecker/) needs RGB converted to hex but has been there for me for over two decades now, heh.

PR #316, opened 2025-03-09
:zap: Debounce search input to improve typing perf

foysalit:
Didn't realize searching perf is so bad in prod. This PR debounces URL update to ensure search input feels smooth.

### DIFF
--- a/components/config/Member.tsx
+++ b/components/config/Member.tsx
@@ -14,13 +14,14 @@ import { createTeamPageLink } from '@/team/helpers'
 import { useRouter, useSearchParams } from 'next/navigation'
 import { Input } from '@/common/forms'
 import { useDebounce } from 'react-use'
+import { useSyncedState } from '@/lib/useSyncedState'
 
 // Make sure we don't update the url query param on every key stroke
 const MemberSearchInput = () => {
   const router = useRouter()
   const searchParams = useSearchParams()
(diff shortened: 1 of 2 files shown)


components/config/Member.tsx
```diff
@@ -13,9 +13,49 @@ import { MagnifyingGlassIcon } from '@heroicons/react/24/solid'
...
 import { Input } from '@/common/forms'
+import { useDebounce } from 'react-use'
 
-export function MemberConfig() {
+// Make sure we don't update the url query param on every key stroke
+const MemberSearchInput = () => {
   const router = useRouter()
+  const searchParams = useSearchParams()
+  const searchQuery = searchParams.get('search') || ''
+  const [inputValue, setInputValue] = useState(searchQuery)
```
devinivy, line 23: One thing about doing it this way is that `inputValue` (which is what displays to the user) does not sync-up with `searchQuery` when the search param changes.

I think what we want in this case is for `inputValue` to sync with `searchQuery` if it ever changes.  But also to be able to deviate from `searchQuery` while debouncing.

This is a super common use case and we have a hook `useSyncedState()` in the project which does this, could be used like:
```suggestion
  const searchQuery = searchParams.get('search') || ''
  const [inputValue, setInputValue] = useSyncedState(searchQuery)
```

foysalit: oh great call!

PR #320, opened 2025-03-24
❇️ Use subject view for workspace data and show warnings

foysalit:
Depends on https://github.com/bluesky-social/atproto/pull/3651
This PR refactors workspace data loading to use `tools.ozone.moderation.getSubjects` endpoint and adds a more advanced filtering system for workspace items that allows grouping and using OR filters between groups. 

Thanks to the unified dataloader, workspace can now also filter by follower count and other profile info and shows warnings when workspace has high profile accounts in it.

Screenshot 2025-03-24 at 22 22 17

> High profile warning in workspace action form

Screenshot 2025-03-24 at 21 54 28

> Advanced grouped filtering for workspace items

Few little touches here and there
1. show star icon in workspace item rows for items of high profile account
2. change label prefix `report:` to a flag icon to save real-estate and make workspace items more legible
3. change `lang:und` tag to `lang?` to save real-estate and add more descriptive title

### DIFF
--- a/app/actions/ModActionPanel/QuickAction.tsx
+++ b/app/actions/ModActionPanel/QuickAction.tsx
@@ -57,6 +57,8 @@ import { EmailComposer } from 'components/email/Composer'
 import { ActionPolicySelector } from '@/reports/ModerationForm/ActionPolicySelector'
 import { PriorityScore } from '@/subject/PriorityScore'
 import { getEventFromFormData } from '@/mod-event/helpers/emitEvent'
+import { Alert } from '@/common/Alert'
+import { TextWithLinks } from '@/common/TextWithLinks'
 
 const FORM_ID = 'mod-action-panel'
 const useBreakpoint = createBreakpoint({ xs: 340, sm: 640 })
@@ -542,7 +544,10 @@ function Form(
                 >
                   {!isSubjectDid && record?.repo && (
                     <div className="-ml-1 my-2">
-                      <RecordAuthorStatus repo={record.repo} />
+                      <RecordAuthorStatus
+                        repo={record.repo}
+                        profile={profile}
+                      />
                     </div>
                   )}
                 </PreviewCard>
@@ -560,9 +565,13 @@ function Form(
                     <LastReviewedTimestamp subjectStatus={subjectStatus} />
                   </p>
                   {!!subjectStatus.comment && (
-                    <Card hint="important" className="mt-2">
-                      <strong>Note:</strong> {subjectStatus.comment}
-                    </Card>
+                    <div className="mt-2">
+                      <Alert
+                        type="info"
+                        title="Note"
+                        body={<TextWithLinks text={subjectStatus.comment} />}
+                      />
+                    </div>
                   )}
                 </div>
               )}
--- a/components/common/TextWithLinks.tsx
+++ b/components/common/TextWithLinks.tsx
@@ -1,7 +1,7 @@
 // Utility function to detect and replace links with <a> tags
 const wrapLinksInText = (text: string): JSX.Element[] => {
   // Regular expression to match URLs
-  const urlRegex = /(https?:\/\/[^\s]+)/g
+  const urlRegex = /(https?:\/\/[^\s]+)/
 
   // Split text into parts, with URLs as matches
   const parts = text.split(urlRegex)
--- a/components/common/TextWithLinks.tsx
+++ b/components/common/TextWithLinks.tsx
@@ -1,0 +1,31 @@
+// Utility function to detect and replace links with <a> tags
+const wrapLinksInText = (text: string): JSX.Element[] => {
+  // Regular expression to match URLs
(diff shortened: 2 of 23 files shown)


app/actions/ModActionPanel/QuickAction.tsx
```diff
@@ -560,9 +565,13 @@ function Form(
                     <LastReviewedTimestamp subjectStatus={subjectStatus} />
                   </p>
                   {!!subjectStatus.comment && (
-                    <Card hint="important" className="mt-2">
-                      <strong>Note:</strong> {subjectStatus.comment}
-                    </Card>
+                    <div className="mt-2">
```
foysalit, line 568: kinda unrelated but mods wanted to give comments more alarming look since we mostly use these comments for critical instruction to other mods.

components/common/TextWithLinks.tsx
```diff
@@ -0,0 +1,31 @@
+// Utility function to detect and replace links with <a> tags
+const wrapLinksInText = (text: string): JSX.Element[] => {
```
foysalit, line 2: Just extracting an already existing component into a dedicated file.

components/tags/SubjectTag.tsx
```diff
@@ -25,6 +26,19 @@ export const SubjectTag = ({
       )
     }
   }
-  
+
+  if (tag.startsWith('report:')) {
```
foysalit, line 30: Unrelated to the PR but workspace items are super cramped for space nowadays due to multiple tags per item with report type. The icon display saves us some real estate while not losing any meaning.

components/common/TextWithLinks.tsx
```diff
@@ -0,0 +1,31 @@
...
+  // Regular expression to match URLs
+  const urlRegex = /(https?:\/\/[^\s]+)/g
+
+  // Split text into parts, with URLs as matches
+  const parts = text.split(urlRegex)
+
+  return parts.map((part, index) => {
+    if (urlRegex.test(part)) {
```
devinivy, line 10: Gotta be a little careful with this because `regexp.test()` is stateful when `regexp` has the global `/g` flag.  Good example here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test#using_test_on_a_regex_with_the_global_flag

Though I don't think the global flag is necessary in here, so may be able to just get rid of it.
```suggestion
  const urlRegex = /(https?:\/\/[^\s]+)/

  // Split text into parts, with URLs as matches
  const parts = text.split(urlRegex)

  return parts.map((part, index) => {
    if (urlRegex.test(part)) {
```

components/workspace/FilterContext.tsx
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  const removeGroup = (groupId: number) => {
+    setFilterGroup((prev) => prev.filter((_, i) => i !== groupId))
+  }
+
+  const removeFilter = (groupId: number, field: string, operator: string) => {
+    setFilterGroup((prev) => {
+      const newGroup = [...prev]
+      newGroup[groupId].filters = newGroup[groupId].filters.filter(
```
devinivy, line 120: I think you're cloning `prev` to avoid mutating, but `prev[groupId]` is getting mutated, appearing both in `prev` and `newGroup`.

components/workspace/FilterContext.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+  const addFilter = (groupId: number, filter: WorkspaceFilterItem) => {
+    setFilterGroup((prev) => {
+      const newGroup = [...prev]
+      if (
+        !newGroup[groupId]?.filters.some(
+          (f) => f.field === filter.field && filter.operator === f.operator,
+        )
+      ) {
+        newGroup[groupId].filters.push(filter)
```
devinivy, line 103: Same here re: mutation.

components/workspace/utils.ts
```diff
@@ -59,53 +62,100 @@ export const buildItemsSummary = (groupedItems: GroupedSubjects): string => {
...
+        ? isRecordCreatedAfterFilter
+        : !isRecordCreatedAfterFilter
+    case 'imageEmbed':
+      return !!data.status?.tags?.includes(EmbedTypes.Image)
+    case 'videoEmbed':
+      return !!data.status?.tags?.includes(EmbedTypes.Video)
+    case 'content':
+      return matchText(
+        filter.value,
+        data.record?.value?.text ? `${data.record?.value?.text}` : undefined,
```
devinivy, line 137:
```suggestion
        data.record?.value?.text ? `${data.record.value.text}` : undefined,
```

devinivy (approved):
Left a few notes, generally looking good 👍

components/workspace/FilterContext.tsx
foysalit: ohh nice catch!

PR #326, opened 2025-04-07
⬆️ Upgrade nextjs to v15

foysalit:
This PR upgrades next.js to v15 and as requirement
1. upgrades react to v19
2. upgrades react-toastify lib to v11
3. upgrades headlessui to v2
4. upgrades react-md-editor to v4
5. upgrades github actions to use latest yarn version


components/subject/Summary.tsx
```diff
@@ -8,7 +8,7 @@ import {
   ShieldExclamationIcon,
   UserCircleIcon,
 } from '@heroicons/react/24/solid'
-import { ComponentProps, useState } from 'react'
+import { ComponentProps, useState, type JSX } from 'react';
```
devinivy, line 11: The semicolon here is out of the project style.  I think this applies in many place, will just leave this one comment though.

app/repositories/[id]/page.tsx
```diff
@@ -1,13 +1,14 @@
...
-  params,
-}: {
-  params: { id: string }
-}) {
+export default function RepositoryViewPage(
+  props: {
+    params: Promise<{ id: string }>
+  }
+) {
+  const params = use(props.params);
```
Comment (devinivy) [inline app/repositories/[id]/page.tsx:11]: Is the promise here a next 15 thing?

app/actions/ModActionPanel/QuickAction.tsx
```diff
@@ -398,7 +398,10 @@ function Form(
         ToolsOzoneModerationDefs.isModEventLabel(coreEvent)
 
       setModEventType(
-        eventMayNeedEmail && !shouldMoveToNextSubject && canSendEmail
+        eventMayNeedEmail &&
+          !shouldMoveToNextSubject &&
+          canSendEmail &&
+          isSubjectDid
```
devinivy, line 404: Is adding `isSubjectDid` as a condition a misc bugfix?

.github/workflows/build.yml
```diff
@@ -12,14 +12,24 @@ jobs:
...
       - name: Install node 20
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
         with:
           node-version: 20
+
+      - name: Enable and use Corepack for Yarn
+        run: |
+          corepack enable
+          corepack prepare yarn@4.8.1 --activate
```
devinivy, line 22: I wonder if placing the yarn version in package.json might be nice, could be picked-up in local dev.

devinivy (approved):

app/repositories/[id]/page.tsx
Comment (foysalit) [reply app/repositories/[id]/page.tsx:11]: yep https://nextjs.org/docs/app/building-your-application/routing/dynamic-routes#example

components/subject/Summary.tsx
foysalit: oh right, codemod applied these. cleaning up now thanks!

app/actions/ModActionPanel/QuickAction.tsx
foysalit: yeah this is a fix for the case where moving to email action on a record subject doesn't really work since the subject needs to be a did/account.

.github/workflows/build.yml
foysalit: yep, set up the `packageManager` field which should take care of it.

PR #330, opened 2025-04-19
:sparkles: Support for verification feature.

foysalit:
This PR adds verification feature support to ozone UI.

A few bigger pieces in here are 

1. verification based filtering in workspace
2. granting and revoking verifications
3. viewing verification history
4. verification badge in all places where applicable

Screenshot 2025-04-19 at 23 03 14

> Workspace verification action popup

Screenshot 2025-04-18 at 16 56 31

> Action panel verified/verifier user warning

Screenshot 2025-04-21 at 21 59 01

> Verification badge in profile card

Screenshot 2025-04-21 at 21 59 18

> Verifier badge in profile card

Screenshot 2025-04-21 at 22 00 10

> Verifier badge in repo page header

Screenshot 2025-04-21 at 22 00 45

> Verified user badge in workspace item

Screenshot 2025-04-21 at 22 11 34

> Verification page

### DIFF
--- a/components/mod-event/SelectorButton.tsx
+++ b/components/mod-event/SelectorButton.tsx
@@ -4,7 +4,11 @@ import { ChevronDownIcon } from '@heroicons/react/20/solid'
 import { Dropdown } from '@/common/Dropdown'
 import { MOD_EVENTS } from './constants'
 import { isReporterMuted, isSubjectMuted } from '@/subject/helpers'
-import { DM_DISABLE_TAG, VIDEO_UPLOAD_DISABLE_TAG } from '@/lib/constants'
+import {
+  DM_DISABLE_TAG,
+  TRUSTED_VERIFIER_TAG,
+  VIDEO_UPLOAD_DISABLE_TAG,
+} from '@/lib/constants'
 import { usePermission } from '@/shell/ConfigurationContext'
 
 const actions = [
@@ -61,6 +65,14 @@ const actions = [
     key: MOD_EVENTS.ENABLE_VIDEO_UPLOAD,
   },
   {
+    text: 'Make Trusted Verifier',
+    key: MOD_EVENTS.MAKE_VERIFIER,
+  },
+  {
+    text: 'Revoke Trusted Verifier',
+    key: MOD_EVENTS.REVOKE_VERIFIER,
+  },
+  {
     text: 'Set Priority Score',
     key: MOD_EVENTS.SET_PRIORITY,
   },
@@ -89,14 +101,8 @@ export const ModEventSelectorButton = ({
   const canTakedown = usePermission('canTakedown')
   const canManageChat = usePermission('canManageChat')
   const canSendEmail = usePermission('canSendEmail')
-  const {
-    takendown,
-    muteUntil,
-    muteReportingUntil,
-    reviewState,
-    appealed,
-    tags,
-  } = subjectStatus || {}
+  const canVerify = usePermission('canVerify')
+  const { takendown, reviewState, appealed, tags } = subjectStatus || {}
   const isMutedSubject = isSubjectMuted(subjectStatus)
   const isMutedReporter = isReporterMuted(subjectStatus)
 
@@ -207,6 +213,19 @@ export const ModEventSelectorButton = ({
         return false
       }
 
+      if (
+        key === MOD_EVENTS.MAKE_VERIFIER &&
+        (tags?.includes(TRUSTED_VERIFIER_TAG) || !isSubjectDid || !canVerify)
+      ) {
+        return false
+      }
+      if (
+        key === MOD_EVENTS.REVOKE_VERIFIER &&
+        (!tags?.includes(TRUSTED_VERIFIER_TAG) || !isSubjectDid || !canVerify)
+      ) {
+        return false
+      }
+
       return true
     })
   }, [
@@ -222,6 +241,7 @@ export const ModEventSelectorButton = ({
     canTakedown,
     canDivertBlob,
     canSendEmail,
+    canVerify,
     forceDisplayActions,
   ])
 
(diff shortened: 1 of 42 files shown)


components/mod-event/SelectorButton.tsx
```diff
@@ -207,6 +213,19 @@ export const ModEventSelectorButton = ({
         return false
       }
 
+      if (
+        key === MOD_EVENTS.MAKE_VERIFIER &&
+        (tags?.includes(TRUSTED_VERIFIER_TAG) || !isSubjectDid || !canVerify)
```
devinivy, line 218: I forget—is there also logic on the server to ensure the user is permitted to add/remove trusted verifier tags?

next.config.js
```diff
@@ -1,9 +1,6 @@
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   reactStrictMode: true,
-  experimental: {
-    missingSuspenseWithCSRBailout: false,
-  },
```
devinivy, line 6: Stood-out, just noting in case it was unintentional.

lib/constants.ts
```diff
@@ -13,6 +13,7 @@ export const PLC_DIRECTORY_URL =
     : 'https://plc.directory')
 
 export const QUEUE_CONFIG = process.env.NEXT_PUBLIC_QUEUE_CONFIG || '{}'
+export const OPT_IN_FEATURES = process.env.NEXT_PUBLIC_OPT_IN_FEATURES || '{}'
```
devinivy, line 16: I didn't see this getting use anywhere—perhaps we can remove it.

components/mod-event/SelectorButton.tsx
foysalit: yep. we do this through dynamically configured protected-tags. this is already set up properly on our instance.

next.config.js
foysalit: yep. this is no longer needed and actually throws a warning during the build.

lib/constants.ts
foysalit: oh right yeah added this before i had the verifier config piece on the backend.

components/verification/ActionButton.tsx
```diff
@@ -0,0 +1,224 @@
...
+        (v) => v.issuer == verifierDid,
+      )
+      if (verification) {
+        revokeUris.push(verification.uri)
+      } else {
+        grantVerifications.push({
+          subject: sub.profile.did,
+          handle: sub.profile.handle,
+          displayName: sub.profile.displayName || '',
+        })
```
devinivy, line 42: Does the backend reject/fail the verification if one already exists?  It's pretty easy to publish two verifications for the same user by mistake.  So having a backstop on ozone backend to help prevent that would be nice if it doesn't already exist.

components/verification/ActionButton.tsx
```diff
@@ -0,0 +1,224 @@
...
+  const revokeUris: string[] = []
+
+  Object.values(subjects || {}).map((sub) => {
+    if (isValidProfileViewDetailed(sub.profile)) {
+      const verification = sub.profile.verification?.verifications?.find(
+        (v) => v.issuer == verifierDid,
+      )
+      if (verification) {
+        revokeUris.push(verification.uri)
+      } else {
```
devinivy, line 37: If there happen to be multiple verifications on a single subject from our verifier (i.e. may have happened in error) it would be neat if this was able to clean them all up.

components/verification/ActionButton.tsx
```diff
@@ -0,0 +1,224 @@
...
+  profile,
+}: {
+  did: string
+  profile: AppBskyActorDefs.ProfileViewDetailed
+}) => {
+  const { verifierDid } = useServerConfig()
+  // Revocation is only allowed when there is a verification record by the issuer that is configured with the ozone agent
+  const revocableVerificationUri = profile.verification?.verifications?.find(
+    (v) => v.issuer == verifierDid,
+  )?.uri
```
devinivy, line 79: Same here—it would be sort of nice if this just acted on any verifications by our verifier, including if there are somehow multiple.

components/verification/ActionButton.tsx
foysalit: no we don't check existing verifications before creating new ones in the backend. frontend tries to make sure we don't show `grant verification` ui if one already exists but yeah, i can add the check on the backend too.

components/verification/FilterPanel.tsx
```diff
@@ -0,0 +1,180 @@
...
+  useEffect(() => {
+    setFormState({
+      subjectsInput: (filters.subjects || []).join(', '),
+      issuersInput: (filters.issuers || []).join(', '),
+      isRevoked:
+        filters.isRevoked === undefined ? 'any' : filters.isRevoked.toString(),
+      createdAfter: filters.createdAfter || '',
+      createdBefore: filters.createdBefore || '',
+    })
+  }, [filters, setFormState])
```
devinivy, line 37: I think conventionally react lint rules don't require you pass `useState()` setters into effect deps because they are designed to maintain their reference.

devinivy: Also just a note that `filter` prop needs to be careful to maintain its reference otherwise it will clobber local `formState` any time a component using `VerificationFilterPanel` re-renders.

devinivy: Laaaast thing—I think we have a hook `useSyncedState()` which implements the pattern used here.

components/verification/ActionButton.tsx
devinivy: Yeah, it's just that the UI is more likely to be out of sync with the actual state.  And nothing stops you from publishing multiple verification records for the same user.

devinivy (approved):

components/verification/FilterPanel.tsx
foysalit: hmm... the useSyncedState thing caused looped updates because the values are derived from filters object so had to revert that back. im sure i can simplify it in some way to make this a bit nicer.

PR #340, opened 2025-05-07
Fix the build for GHCR

trezy:
The Docker build has been failing since [09 April](https://github.com/bluesky-social/ozone/actions/runs/14331039613). This means anybody relying on the container from GHCR (including any users that followed the instructions in [HOSTING.md](HOSTING.md)) have been stuck on Ozone v0.1.84.

I was able to reproduce the issue locally by running `docker buildx build .` without installing any dependencies. Installing dependencies before running `docker buildx build .` resolves the issue.

### DIFF
--- a/.github/workflows/build-and-push-ghcr.yaml
+++ b/.github/workflows/build-and-push-ghcr.yaml
@@ -25,6 +25,17 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
+      - name: Install node 20
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+
+      - name: Enable Corepack
+        run: corepack enable
+
+      - name: Yarn install
+        run: yarn
+
       - name: Setup Docker buildx
         uses: docker/setup-buildx-action@v2
 


devinivy:
Just plopping a reference here for a similar change we had to make on the PDS distro https://github.com/bluesky-social/pds/commit/d0c43fbd4f70dbe307597c83b1c9020988978dd4

trezy:
@devinivy Yeah, I tried to keep the changes as minimal as possible. I did consider adding the `packageManager` field (as well as a `.yarnrc.yml`) to the `service` directory, but it wasn't necessary to resolve the issue so I left it out. I'm happy to add it if you think it would be helpful.

Also of note, [`corepack prepare` is deprecated](https://github.com/nodejs/corepack/issues/624). Additionally, its functionality (setting the global default package manager version) is only necessary if operating outside of the project. As long as there's a `package.json` that `corepack` can read the `packageManager` field from, it'll do all of the necessary work automatically. Thus, adding `corepack prepare` to this project is unnecessary, and it'd be a good idea to remove it from other projects when those Dockerfiles are touched next.

PR #349, opened 2025-05-26
:bug: Render embedded record along with media when both exists on a post

foysalit:


components/common/posts/PostsFeed.tsx
```diff
@@ -344,6 +376,13 @@ export function PostEmbeds({
       return recordView
     }
   }
+
+  if (AppBskyEmbedRecord.isViewRecord(embed)) {
+    const recordView = <RecordEmbedView embed={{ record: embed }} />
+    if (recordView) {
+      return recordView
+    }
```
devinivy, line 384: `recordView` always exists here, so the conditional is not needed.

devinivy (approved):

PR #350, opened 2025-05-28
:lipstick: Add reply context when reviewing a post which is a reply to a post

foysalit:
Screenshot 2025-05-28 at 16 28 24

> Reply context link

Along with reply context, this PR also improves vertical spacing in the quick action panel and adds account level tag/label display when reviewing a record to provide additional context about the account without having to switch to account view.

Screenshot 2025-05-28 at 16 27 01

> Account level labels and tags display

### DIFF
--- a/components/common/posts/ReplyParent.tsx
+++ b/components/common/posts/ReplyParent.tsx
@@ -4,13 +4,12 @@ import { ReactNode } from 'react'
 import { useActionPanelLink } from '../useActionPanelLink'
 
 export const ReplyParent = ({
-  reply,
+  parent,
   inline = false,
 }: {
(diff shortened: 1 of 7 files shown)


components/common/posts/ReplyParent.tsx
```diff
@@ -59,6 +58,27 @@ export const ReplyParent = ({
...
+        ) : (
+          <>
+            <LinkToPost post={parent.post} /> to an anonymous user
+          </>
+        )}
+      </Wrapper>
+    )
+  }
+
+  console.log('here', parent.$type)
```
devinivy, line 81:
```suggestion
```

PR #365, opened 2025-07-22
:sparkles: Support expanded custom reporting categories

foysalit:
Depends on https://github.com/bluesky-social/atproto/pull/3881

Screenshot 2025-07-22 at 16 47 32

> New reporting types in event filter panel

Screenshot 2025-07-22 at 16 47 59

> New reporting types in event and tag view

Screenshot 2025-07-22 at 16 48 22

> New reporting types in report panel

### DIFF
--- a/components/reports/ReasonBadge.tsx
+++ b/components/reports/ReasonBadge.tsx
@@ -1,5 +1,5 @@
-import { ComAtprotoModerationDefs } from '@atproto/api'
 import { ComponentProps } from 'react'
+import { reasonTypeOptions, groupedReasonTypes } from './helpers/getType'
 
 export function ReasonBadgeButton(
   props: {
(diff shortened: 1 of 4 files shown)


components/reports/ReasonBadge.tsx
```diff
@@ -25,22 +26,57 @@ export function ReasonBadgeButton(
...
+  // Remove namespace prefixes and make more readable
+  return reasonType
+    .replace('com.atproto.moderation.defs#reason', '')
+    .replace('tools.ozone.report.defs#reason', '')
+    .replace(/([A-Z])/g, ' $1')
+    .trim()
+    .toLowerCase()
+    .split(' ')
+    .map(word => word.charAt(0).toUpperCase() + word.slice(1))
+    .join(' ')
```
devinivy, line 81: With string manipulation like this I like to just give an example of the transformation in a comment.

components/reports/ReasonBadge.tsx
```diff
@@ -25,22 +26,57 @@ export function ReasonBadgeButton(
...
+}
+
+// Get the color class for a reason type based on its category
+function getReasonColor(reasonType: string): string {
+  const category = getReasonCategory(reasonType)
+  return groupColors[category] || groupColors.default
+}
+
+// Fallback function to create readable text for unknown reason types
+function getReadableReasonType(reasonType: string): string {
```
devinivy, line 71: I see mappings in `components/reports/helpers/getType.ts`—would this potentially fit better there?

devinivy (approved):

components/reports/ReasonBadge.tsx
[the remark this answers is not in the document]
foysalit: yep yep, totally, i think i added the regex approach later and forgot to take out the split approach.

PR #367, opened 2025-08-04
❇️ Replace event stream view with timeline view in quick action panel

foysalit:


foysalit:
We won't be moving forward with this, at least not any time soon and before we do, we will need more input from mod team on how to provide better but condensed data insight.

PR #371, opened 2025-08-27
:sparkles: Revoke credentials action

foysalit:


components/communication-template/action-template.tsx
```diff
@@ -0,0 +1,195 @@
+import { useState } from 'react'
```
foysalit, line 1: Component that handles ONLY the instance-wide action template config/association, independent of the revokeCredentials action.

lib/util.ts
```diff
@@ -188,3 +188,129 @@ export function isNonNullable<V>(v: V): v is NonNullable<V> {
...
+            results.push({
+              item,
+              success: false,
+              error: error?.message || 'Unknown error'
+            })
+            failed++
+          })
+          break
+        }
+      }
```
devinivy, line 302: The rate limit checks, error handling, and retry logic are sort of sprawling throughout here, I have a feeling there's a way these could be tidied up.  It's not really essential, so I don't think it needs to block the work, but if you see any opportunities in there it could be worth a look.

devinivy (approved):

lib/util.ts
foysalit: yh i dont see us needing this outside of api calls so cleaned up the rate limit stuff. made another pass through and wanted to clarify that a lot of the verbosity comes from error handling and edge case handling and can't see a lot of room for improvements around those.

PR #383, opened 2025-10-04
Update HOSTING.md with security note for PostgreSQL

ThisIsMissEm:
There's been some research showing that ozone installs have been exposing the PostgreSQL database to the public internet, so adding in a more stern warning here to discourage that practice: https://bsky.app/profile/valkyrie.hacker.gf/post/3m2dpzreuuc2t


ThisIsMissEm:
@haileyok sure, that works too!

PR #386, opened 2025-10-06
Bind postgres to localhost

devinivy:
This ensures that postgres binds to localhost rather than to all addresses in the ozone distribution.  It is a secondary measure to help protect against installations that are running without a firewall—it is always recommended to run a firewall and deny traffic on ports not in use by the ozone web server (i.e. 80 and 443).

### DIFF
--- a/service/compose.yaml
+++ b/service/compose.yaml
@@ -32,6 +32,7 @@ services:
   postgres:
     container_name: postgres
     image: postgres:14.11-bookworm
+    command: postgres -c listen_addresses=localhost
     network_mode: host
     restart: unless-stopped
     healthcheck:


haileyok (approved):
yep

ThisIsMissEm:
What about using `-p 127.0.0.1:5432:5432` instead, which in the compose file would be:
```
ports:
  - '127.0.0.1:5432:5432'
```

I think that'd cause it to only bind to localhost, not all interfaces, per:
> `HOST` is `[IP:](port | range)` (optional). If it is not set, it binds to all network interfaces (0.0.0.0).

https://docs.docker.com/reference/compose-file/services/#ports

haileyok:
I think (though @devinivy can correct if wrong) that we'd rather keep `network_mode: host` here. It probably isn't a huge deal with low throughput so performance problems are probably not a real concern here, though I do personally like to keep things in `host` mode.

PR #390, opened 2025-10-10
Add more detailed security information to HOSTING.md

heikadog:
Adds slightly more detailed information about inbound and outbound access. It's not as clear that outbound shouldn't be open to all ports, so this is to mitigate that. Also added port 22 to the list.

Open to refactoring this into a table if needed.

### DIFF
--- a/HOSTING.md
+++ b/HOSTING.md
@@ -21,7 +21,7 @@ Ensure that you can ssh to your server and have root access.
 
 - Public IPv4 address
 - Public DNS name
-- Public inbound and outbound internet access permitted on port 80/tcp, 443/tcp, and 22/ssh
+- Public inbound and outbound internet access permitted on port 80/tcp, 443/tcp, 53/tcp, and 22/ssh
 
 **Server Recommendations**
 


HOSTING.md
```diff
@@ -40,10 +40,12 @@ Ensure that you can ssh to your server and have root access.
 
 One of the most common sources of misconfiguration is not opening firewall ports correctly. Please be sure to double check this step.
 
-In your cloud provider's console, the following ports should be open to inbound access from the public internet.
+In your cloud provider's console, the following ports should be open to inbound and outbound access to and from the public internet.
```
bnewbold: I agree with this inbound/outbound clarification.

HOSTING.md
```diff
@@ -40,10 +40,12 @@ Ensure that you can ssh to your server and have root access.
...
 
 - 80/tcp (Used only for TLS certification verification)
 - 443/tcp (Used for all application requests)
+- 53/tcp (Used for domain name resolution)
```
bnewbold: Is this specifically needed for DNS? I think we can just rely on default system config, which often uses the system resolver on the server, which is often auto-configured to talk to the hosting provider's DNS server over UDP.

heikadog: loading the queue didn't work until i added it in prod and the network requests seemed to also align with that !

PR #396, opened 2025-10-28
workspace: check for did or uri in all csv cols if no header found

haileyok:
Lots of places that we export data from (i.e. Osprey) don't have column headers that exactly match `did` or `uri`. For example, we might export something that has the header `SubjectDid` instead.

This PR would continue to support the `did` or `uri` header value as the "default", but then fallback to trying to find any column with values matching either a did or an at uri, then use that column for the workspace import instead.

### DIFF
--- a/lib/csv.ts
+++ b/lib/csv.ts
@@ -1,3 +1,4 @@
+import { isDid } from '@atproto/api'
 import { parseAtUri } from './util'
 
 export type CsvContent = {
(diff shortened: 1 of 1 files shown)


lib/csv.ts
```diff
@@ -97,9 +99,23 @@ export const extractFromCSV = (data: string): string[] => {
...
   const headers = header.split(',').map(cleanCSVColumn)
-  const didIndex = headers.indexOf('did')
-  const uriIndex = headers.indexOf('uri')
+  let didIndex = headers.indexOf('did')
+  let uriIndex = headers.indexOf('uri')
+
+  // before returning an empty array due to no headers, lets first try and parse out a did or an aturi from the data
+  if (didIndex === -1 && uriIndex === -1 && content.length >= 1) {
+    for (const [idx, col] of content.entries()) {
+      if (col.startsWith('did:')) {
```
haileyok: embarrassed to admit that i don't actually know what the sdk function is for parsing a did/uri lol (or if one exists). probably a better way to do these

ThisIsMissEm: I think you'd be looking for `isDid` from `@atproto/api`

ThisIsMissEm: Well, `@atproto/api`'s code has an isDid method but it's not exported. So instead from `@atproto/did`

Issue #280, opened 2025-01-22
applying/removing a label with the same identifier as an external labeler doesn't work


derekvanvliet:
I'm subscribed to an external labeler that has a label with the same identifier as one that my labeler uses. I'm finding that if I try to add or remove that label from a piece of content that the other labeler's label is on, it reports that the content was labeled, but it does not appear on the content.

PR #452, opened 2026-02-22
Deduplicate labels on negation to prevent internal server error

aitorres:
## Related issues

Addresses #280

## Bug

If an account has been applied two labels with the same value `foo` by two different labelers (e.g. our own labeler vs. an external labeler `bar`), trying to remove our own `foo` results in an Internal Server Error since the front-end tries to emit a record negating the same label value twice (`['foo', 'foo']`

## Fix

Deduplicate current values to handle this case. If the account has the same label value applied multiple times, we'll only count it once when figuring out which ones to negate.

## Validation

Tested on an Ozone self-hosted instance, this gets rid of the issue and allows the label to be removed.

### DIFF
--- a/app/actions/ModActionPanel/useQuickAction.tsx
+++ b/app/actions/ModActionPanel/useQuickAction.tsx
@@ -18,6 +18,7 @@ import {
   HOUR,
   pluralize,
   takesKeyboardEvt,
+  unique,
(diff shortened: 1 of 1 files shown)


foysalit (approved):
Looks good, thank you!

PR #508, opened 2026-06-01
fix: report escalated actions

gcwill70:
Fix bug where events that were escalated to account level had invalid `reportAction` added to them

### DIFF
--- a/app/reports/[id]/page-content.tsx
+++ b/app/reports/[id]/page-content.tsx
@@ -14,16 +14,9 @@ import { LabelSelector } from '@/common/labels/Selector'
 import { isSelfLabel } from '@/common/labels/util'
 import { ConfirmationModal } from '@/common/modals/confirmation'
 import { useWorkspaceOpener } from '@/common/useWorkspaceOpener'
-import { MessageActorMeta } from '@/dms/MessageActorMeta'
-import { ModEventDetailsPopover } from '@/mod-event/DetailsPopover'
-import { LastReviewedTimestamp } from '@/subject/LastReviewedTimestamp'
-import { RecordAuthorStatus } from '@/subject/RecordAuthorStatus'
-import { SubjectTag } from 'components/tags/SubjectTag'
-import { HighProfileWarning } from '@/repositories/HighProfileWarning'
 import { ConversationSubjectWarning } from '@/dms/ConversationSubjectWarning'
-import { PriorityScore } from '@/subject/PriorityScore'
-import { Alert } from '@/common/Alert'
-import { TextWithLinks } from '@/common/TextWithLinks'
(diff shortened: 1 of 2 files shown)


app/reports/[id]/page-content.tsx
```diff
@@ -546,31 +552,74 @@ function ReportDetailLayout(props: {
       }
     }
 
-    const eventType = (finalVals.event as any)?.$type as string | undefined
+    const subj = (finalVals as any).subject
+    const event = finalVals.event
+    const eventType = event?.$type as string | undefined
+    // check if event was escalated to the owning account
+    const isEscalated =
```
Comment (foysalit) [inline app/reports/[id]/page-content.tsx]: wonder if the `escalated` term may be a bit confusing here since `escalate` means an entirely different action in the user flow? `isPropagatedToAccount` sounds too technical but does convey exactly what's happening.

app/reports/[id]/page-content.tsx
```diff
@@ -546,31 +552,74 @@ function ReportDetailLayout(props: {
...
+          await onSubmit({
+            ...finalVals,
+            event: {
+              ...event,
+              comment:
+                `[ESCALATION_ACTION]: This action was taken after actioning this report: \n${reportUrl}\n\n${event.comment || ''}`.trim(),
+            },
+          })
+        } else {
+          await onSubmit(finalVals)
```
Comment (foysalit) [inline app/reports/[id]/page-content.tsx:587]: sorry, im probably not seeing the full picture here but it seems like we are calling onSubmit twice if there is no comment in the event but only once if there is a comment?

foysalit (changes requested):
want to understand the implementation a bit better.

app/reports/[id]/page-content.tsx
Comment (gcwill70) [reply app/reports/[id]/page-content.tsx]: Yeah I couldn't think of a better word. Re-worded to `ELEVATED_ACTION`. wdyt? Is propagate better?

app/reports/[id]/page-content.tsx
Comment (gcwill70) [reply app/reports/[id]/page-content.tsx:587]: I traced `onSubmit` and it calls `emitEvent` and a few query invalidations. So calling it twice shouldn't have harmful side effects.

This was a best-effort way to minimize emitted events. It will add to the comment if possible and if not, it will emit the original action and a comment event. (Most if not all "elevated" actions we support have a comment field so only 1 event is emitted.)

Maybe there a better place I could put the elevation comment? Perhaps in `modTool`?

Screenshot 2026-06-02 at 9 41 47 AM

app/reports/[id]/page-content.tsx
Comment (gcwill70) [reply app/reports/[id]/page-content.tsx]: reworded to `CASCADED` after internal discussion

gcwill70:
Closing this. Going to update backend instead to allow cascaded actions

PR #511, opened 2026-06-04
Various Report-level Fixes

gcwill70:
* Handle actions that cascade onto the account by creating a report note
* Add online moderators indicator
* Add link to open accounts in action panel from report page
* Create `useReports` hook to manage report viewing and navigation
* Add 'advance after close' to report page
* Update report reason types
* Fix flaky tests

### DIFF
--- a/components/reports/ReportActions.tsx
+++ b/components/reports/ReportActions.tsx
@@ -98,7 +98,6 @@ function TransitionConfirmPanel({
 }) {
   const [note, setNote] = useState('')
   const createActivity = useCreateActivity()
-  const { advanceToNext } = useReports(reportId)
   const { activityType, confirmLabel } = ACTION_CONFIG[action]
 
   const handleConfirm = () => {
@@ -117,9 +116,8 @@ function TransitionConfirmPanel({
           if (action === 'no-action' && onResolveAppeal) {
             await onResolveAppeal()
           }
-          if (action === 'no-action') {
-            advanceToNext()
-          }
+          // Auto-advance is handled centrally by useAdvanceOnReportClose,
+          // which fires when the report's status transitions to 'closed'.
           onDone()
         },
         onError: (e) => {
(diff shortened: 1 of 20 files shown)


components/team/useOnlineModerators.ts
```diff
@@ -0,0 +1,92 @@
...
+export type OnlineModerator = {
+  did: string
+  lastActive: Date
+}
+
+const ACTIVE_THRESHOLD = 5 * 60 * 1000
+const REFETCH_INTERVAL = 5 * 1000
+const STALE_TIME = 2 * 1000 // debounce protection when used with refetch below
+
+export function useOnlineModerators() {
```
foysalit, line 34: this is pretty cool! wonder if we should cross match this with member list just so dids from automod accounts dont show up?

components/team/useOnlineModerators.ts
```diff
@@ -0,0 +1,92 @@
...
+      const update = (did: string, activityDate: Date) => {
+        const existing = moderatorActivity.get(did)
+        if (!existing || activityDate > existing) {
+          moderatorActivity.set(did, activityDate)
+        }
+      }
+
+      // mod events
+      try {
+        const { data } = await labelerAgent.tools.ozone.moderation.queryEvents({
```
foysalit, line 56: we should probably limit these by action types and only limit to like takedown, label, email etc. actions?  doesnt make sense to include report actions for instance.

components/reports/ReportActions.tsx
```diff
@@ -320,6 +366,13 @@ export function ReportActionsBar({
         </button>
       </div>
 
+      <Checkbox
```
foysalit, line 369: this is pretty nice UX! i do wonder though, if we should start thinking of these as user level configs/preferences that can be managed in one place. just something to think about in the future.

foysalit (approved):
Looks great thanks!

components/team/useOnlineModerators.ts
gcwill70: Good catch! Fixed

components/team/useOnlineModerators.ts
gcwill70: Agreed! Fixed

PR #513, opened 2026-06-10
Add conversation support to report level features

gcwill70:
- Conversation checkbox in the queue create form
- Conversation option in the queue list subject filter
- Message and Conversation options in the beta report list filter; clear collections for all non-record subject types
- Cypress coverage for creating a conversation queue

Dependent on https://github.com/bluesky-social/atproto/pull/5076

### DIFF
--- a/app/reports/beta/Filters.tsx
+++ b/app/reports/beta/Filters.tsx
@@ -34,6 +34,8 @@ const SUBJECT_TYPE_OPTIONS = [
   { id: 'all', text: 'All Subject Types', value: '' },
   { id: 'account', text: 'Account', value: 'account' },
   { id: 'record', text: 'Record', value: 'record' },
+  { id: 'message', text: 'Message', value: 'message' },
+  { id: 'conversation', text: 'Conversation', value: 'conversation' },
 ]
 
 function useFiltersUpdater() {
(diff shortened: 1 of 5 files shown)


app/reports/beta/Filters.tsx
```diff
@@ -91,10 +93,11 @@ export function BetaReportsFilters() {
   const setSubjectType = (value: string) => {
     if (value === subjectType) {
       updateParam({ subjectType: undefined, collections: undefined })
-    } else if (value === 'account') {
-      updateParam({ subjectType: 'account', collections: undefined })
+    } else if (value === 'record') {
```
foysalit, line 96: sorry im not sure i understand why this change is needed?

foysalit (approved):

app/reports/beta/Filters.tsx
gcwill70: As part of updating subjectTypes to have a new `conversation` value, I saw this needed some fixing.

`record` is the only subject type that can have a `collections` field so we handle it here. All other subject types should clear out `collections` (`account`, `message`, `conversation`)

foysalit (approved):
nice!

PR #527, opened 2026-07-07
Image search

gcwill70:
Adds image searching tool. Will show link in sidebar if url is set in env

### DIFF
--- a/app/api/image-search/enabled/route.ts
+++ b/app/api/image-search/enabled/route.ts
@@ -1,0 +1,6 @@
+// Reports whether image search is configured, without exposing the backend
+// URL (which is server-only env). Clients use this to decide whether to show
+// image-search UI; the actual searches go through the sibling proxy route.
+export async function GET() {
+  return Response.json({ enabled: !!process.env.IMAGE_SEARCH_API_URL })
+}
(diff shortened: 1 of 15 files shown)


app/api/image-search/enabled/route.ts
```diff
@@ -0,0 +1,6 @@
+// Reports whether image search is configured, without exposing the backend
+// URL (which is server-only env). Clients use this to decide whether to show
+// image-search UI; the actual searches go through the sibling proxy route.
+export async function GET() {
```
foysalit: we do have a `useServerConfig` hook which loads initial server config along with roles, permissions etc. might be nice to inject this in there to avoid proliferating endpoints to fetch config.

app/api/image-search/search/route.ts
```diff
@@ -0,0 +1,116 @@
...
+  const hash = searchParams.get('hash')
+  const url = searchParams.get('url')
+  if (!hash === !url) {
+    return Response.json(
+      { error: 'Need either a hash or url' },
+      { status: 400 },
+    )
+  }
+
+  try {
```
foysalit, line 83: wonder if we need better typing for responses here?

foysalit (approved):
looking good. just a few comments regarding code quality but feature looks great!

app/api/image-search/enabled/route.ts
gcwill70: Sounds good. I mad the url readable in the browser and populate it in useServerConfig.

Let me know if you'd rather update `tools.ozone.server.getConfig` to return the url

PR #532, opened 2026-07-16
:bug: various UI fixes

gcwill70:
### Image search: configurable default lookback (default 14 days)
The lookback default on the image search page was hardcoded to 7 days. It is now configurable via `NEXT_PUBLIC_IMAGE_SEARCH_DEFAULT_LOOKBACK_DAYS` and defaults to 14 when unset. The `lookbackDays` URL param still overrides it as before.

### Queue cards: fix overflow when many moderators are assigned
Queues with many assignees blew out the card layout since every assignee chip rendered inline. The assigned list now follows the same convention as the Filters summary on the card: a compact `Assigned: N moderators` pill with the full `QueueAssigneeStatus` (assignee chips, assign/unassign actions) revealed in a `Hover` panel.

## Test plan
- `yarn type-check` passes
- Verified no cypress specs reference the `queue-assignees` selector (it is preserved inside the hover panel regardless)


foysalit (approved):
looks good thanks!

PR #570, opened 2026-08-26
fix: persist report order

gcwill70:
Store report ordering so that report queue ordering survives page loads


foysalit (approved):
wonder if we need a prominent indicator in the UI to show which way the list is being sorted?

PR #572, opened 2026-09-04
Add account preferences to repository details

foysalit:
## Summary

- add a read-only User Preferences section to repository details above Account History
- lazily request account preferences only when the collapsed section is expanded
- render saved feeds, content filters, feed controls, muted words, interests, moderation services, and hidden posts in shape-specific tables
- upgrade `@atproto/api` to 0.20.42 for the current preference lexicon types

The preferences request uses the agent's authenticated fetch handler directly. `@atproto/api` 0.20.42 does not yet bundle the `tools.ozone.moderation.getAccountPreferences` lexicon, so `Agent.call()` rejects the method before making a request. The fetch handler bypasses that lookup while preserving session authentication and Ozone proxy routing.

Depends on bluesky-social/atproto#5421.

## Verification

- `yarn type-check`
- `yarn build`
- `git diff --check`
- adversarial cross-model review: no findings

The production build reports existing lint warnings in unrelated files.

## AI assistance

Implemented and reviewed with Codex/LLM assistance.

### DIFF
--- a/components/repositories/AccountPreferences.tsx
+++ b/components/repositories/AccountPreferences.tsx
@@ -1,0 +1,410 @@
+'use client'
+
+import { AppBskyActorDefs } from '@atproto/api'
+import {
+  ChevronDownIcon,
+  ChevronRightIcon,
+} from '@heroicons/react/20/solid'
+import { useQuery } from '@tanstack/react-query'
+import { type ReactNode, useState } from 'react'
+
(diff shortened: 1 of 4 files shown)


gcwill70 (approved):
Looks good!

We may want to bump @atproto/xrpc so it can resolve to the version @atproto/api requests

